PR #6, opened 2018-10-12
Add aws account to the Lambda runner event

ranrib:
Adding the AWS account ID in the Lambda runner event, for future use.

### DIFF
--- a/src/runners/aws_lambda.js
+++ b/src/runners/aws_lambda.js
@@ -31,10 +31,14 @@ function createRunner(originalContext, resourceType = 'lambda') {
     ]);
 
     runner.setResource(runnerResource);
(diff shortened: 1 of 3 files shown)


src/runners/aws_lambda.js
```diff
@@ -31,10 +31,14 @@ function createRunner(originalContext, resourceType = 'lambda') {
     ]);
 
     runner.setResource(runnerResource);
+    const invokedFunctionArn = (originalContext.invokedFunctionArn) ?
```
galbash: this is not the invokedFunctionArn, this is the AWSAccountNumber

galbash (changes requested):
small change

galbash (approved):

PR #7, opened 2018-10-15
[feat] Add support for pg instrumentation

ilyaLibin:
Node library to support pg npm package. Please focus on `pg.js` file when you review the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,8 @@
     "eslint-plugin-react": "^7.11.0",
     "husky": "^1.1.0",
     "mocha": "^5.0.4",
+    "pg": "^7.5.0",
+    "pg-pool": "^2.0.3",
     "proxyquire": "^2.0.1",
     "rollup": "^0.66.2",
     "rollup-plugin-commonjs": "^9.1.8",
@@ -64,10 +66,13 @@
     "rollup-plugin-json": "^3.1.0",
     "rollup-plugin-terser": "^3.0.0",
     "semantic-release": "^15.9.16",
-    "sinon": "^4.3.0"
+    "sinon": "^4.3.0",
+    "uglify-es": "^3.3.9"
   },
   "peerDependencies": {
-    "aws-sdk": "^2.197.0"
+    "pg": "^7.5.0",
+    "aws-sdk": "^2.197.0",
+    "pg-pool": "^2.0.3"
   },
   "dependencies": {
     "axios": "^0.18.0",
@@ -75,6 +80,7 @@
     "google-protobuf": "^3.5.0",
     "json.sortify": "^2.2.2",
     "md5": "^2.2.1",
+    "node-sqlparser": "^1.0.4",
     "shimmer": "^1.2.0",
     "uuid4": "^1.0.0"
   }
--- a/src/events/pg.js
+++ b/src/events/pg.js
@@ -1,0 +1,117 @@
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const pg = require('pg');
+const Pool = require('pg-pool');
+const sqlParser = require('node-sqlparser');
+const utils = require('../utils.js');
+const tracer = require('../tracer.js');
+const serverlessEvent = require('../proto/event_pb.js');
+const eventInterface = require('../event.js');
+const errorCode = require('../proto/error_code_pb.js');
+
+/**
+ * Wraps the pg's module request function with tracing
+ * @param {Function} wrappedFunction The pg's module
+ * @returns {Function} The wrapped function
+ */
+function pgClientWrapper(wrappedFunction) {
+    return function internalpgClientWrapper(queryString, arg1, arg2) {
+        let patchedCallback;
+        const callback = (arg2 === undefined) ? arg1 : arg2;
+        const params = (arg1 instanceof Function) ? [] : arg1;
+        let sqlObj = {};
+
+        try {
+            sqlObj = sqlParser.parse(queryString);
+        } catch (error) {
+            sqlObj.type = 'Can not process';
+            sqlObj.table = 'Can not process';
+        }
+
+        let databaseType = 'local';
+
+        const {
+            type,
+            table,
+        } = sqlObj;
+
+        const {
+            database,
+            host,
+        } = this.connectionParameters || this._clients[0]; // eslint-disable-line
+
+        if (host.match('.rds.')) { databaseType = 'rds'; }
+        if (host.match('.redshift.')) { databaseType = 'redshift'; }
+
+        try {
+            const resource = new serverlessEvent.Resource([
+                database, // name of the database
+                databaseType,
+                queryString,
+            ]);
+
+            const startTime = Date.now();
+
+            const dbapiEvent = new serverlessEvent.Event([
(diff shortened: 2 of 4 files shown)


src/events/pg.js
```diff
@@ -0,0 +1,101 @@
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const pg = require('pg')
```
galbash: This should be wrapped with a try catch, and if it fails skip this module instead of exploding. Maybe put the required in the init function (called only once anyway)

src/events/pg.js
```diff
@@ -0,0 +1,101 @@
...
+            tracer.addException(error);
+        }
+
+        return wrappedFunction.apply(this, [queryString, params, patchedCallback]);
+    }
+}
+
+module.exports = {
+    /**
+     * Initializes the http tracer
```
galbash: Pg

src/events/pg.js
```diff
@@ -0,0 +1,101 @@
...
+                        callback(res);
+                    }
+                    resolve();
+                };
+            }).catch((err) => {
+                tracer.addException(err);
+            });
+
+            responsePromise.then(() => {
+                dbapiEvent.setDuration(utils.createDurationTimestamp(startTime));
```
ranrib: Is there anything else you can gather from the response? maybe returned count of rows/value?

src/events/pg.js
```diff
@@ -0,0 +1,101 @@
...
+                databaseType,
+                queryString
+            ]);
+
+            const startTime = Date.now();
+
+            const dbapiEvent = new serverlessEvent.Event([
+                `dbapi-${uuid4()}`,
+                utils.createTimestampFromTime(startTime),
+                null,
```
ranrib, line 58: `resource`

src/events/pg.js
```diff
@@ -0,0 +1,101 @@
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const pg = require('pg')
+const sqlParser = require('node-sqlparser');
```
ranrib, line 5: This package is not in the package.json, is something missing?

package.json
```diff
@@ -75,6 +76,7 @@
     "google-protobuf": "^3.5.0",
     "json.sortify": "^2.2.2",
     "md5": "^2.2.1",
+    "pg": "^7.5.0",
```
ranrib: This one should move to peers

src/events/pg.js
ilyaLibin: no

src/events/pg.js
ilyaLibin: You missing nothing. I'm just was stupid :D

package.json
```diff
@@ -57,16 +57,21 @@
     "eslint-plugin-react": "^7.11.0",
     "husky": "^1.1.0",
     "mocha": "^5.0.4",
+    "pg": "^7.5.0",
     "proxyquire": "^2.0.1",
     "rollup": "^0.66.2",
     "rollup-plugin-commonjs": "^9.1.8",
     "rollup-plugin-eslint": "^5.0.0",
     "rollup-plugin-json": "^3.1.0",
+    "rollup-plugin-node-resolve": "^3.4.0",
```
galbash: do you still need this?

package.json
```diff
@@ -57,16 +57,21 @@
...
+    "rollup-plugin-peer-deps-external": "^2.2.0",
```
galbash: same

package.json
```diff
@@ -75,6 +80,7 @@
     "google-protobuf": "^3.5.0",
     "json.sortify": "^2.2.2",
     "md5": "^2.2.1",
+    "node-sqlparser": "^1.0.4",
```
galbash, line 83: check the size of this package to see it is not bloating us

src/events/pg.js
```diff
@@ -0,0 +1,122 @@
...
+const errorCode = require('../proto/error_code_pb.js');
+
+/**
+ * Wraps the pg's module request function with tracing
+ * @param {Function} wrappedFunction The pg's module
+ * @returns {Function} The wrapped function
+ */
+function pgClientWrapper(wrappedFunction) {
+    return function internalpgClientWrapper(queryString, arg1, arg2) {
+        // eslint-disable-next-line no-var
```
galbash: why is this var? try using let

src/events/pg.js
```diff
@@ -0,0 +1,122 @@
...
+        var patchedCallback;
+
+        const callback = (arg2 === undefined) ? arg1 : arg2;
+        const params = (arg1 instanceof Function) ? [] : arg1;
+        let sqlObj = {};
+
+        try {
+            sqlObj = sqlParser.parse(queryString);
+        } catch (error) {
+            tracer.addException(error);
```
galbash: not sure if we should add a tracer exception here (this will explode for every time the user runs an invalid query). lets just make sure we always add the queryString to the event, even if the parsing fails

src/events/pg.js
```diff
@@ -0,0 +1,122 @@
...
+
+            tracer.addEvent(dbapiEvent, responsePromise);
+        } catch (error) {
+            tracer.addException(error);
+        }
+
+        try {
+            return wrappedFunction.apply(this, [queryString, params, patchedCallback]);
+        } catch (error) {
+            throw error;
```
galbash: I don't think this can throw an error (it was the sqlParser thing), lets remove this try / catch (it has no meaning now - catching just to throw)

src/events/pg.js
```diff
@@ -0,0 +1,122 @@
...
+    };
+}
+
+module.exports = {
+    /**
+     * Initializes the pg tracer
+     */
+    init() {
+        shimmer.wrap(pg.Client.prototype, 'query', pgClientWrapper);
+        shimmer.wrap(pg.Pool.prototype, 'query', pgClientWrapper);
```
galbash: did you test with pool as well?

src/patcher.js
```diff
@@ -4,10 +4,11 @@
  */
 const awsSDKPatcher = require('./events/aws_sdk.js');
 const httpPatcher = require('./events/http.js');
+const pgPatcher = require('./events/pg.js');
 const config = require('./config.js');
 
 if (!config.config.isEpsagonPatchDisabled) {
-    [awsSDKPatcher, httpPatcher].forEach((patcher) => {
+    [awsSDKPatcher, httpPatcher, pgPatcher].forEach((patcher) => {
```
galbash, line 18: don't you have to add something to the httpPatcher so it will ignore PG's request? or is it not using it?

galbash (changes requested):
Looks really good, some small comments

src/patcher.js
ilyaLibin: no using http

src/events/pg.js
```diff
@@ -0,0 +1,117 @@
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const pg = require('pg');
```
galbash, line 3: either try catch here or on requiring this module

galbash (changes requested):
only make sure it doesn't explode

galbash (approved):
OK

PR #12, opened 2018-11-14
feat(instrumentation): collecting the return value of the lambda

ronyis:


src/config.js
```diff
@@ -23,6 +23,13 @@ module.exports.config = {
     },
 };
 
+/**
+ * @returns {object} The config object
+ */
+module.exports.getConfig = function getConfig() {
```
galbash, line 29: this is ok, but make the config objects not be module.exports.config but just config

src/wrappers/lambda.js
```diff
@@ -5,11 +5,15 @@ const uuid4 = require('uuid4');
...
 const tracer = require('../tracer.js');
 const utils = require('../utils.js');
+const { getConfig } = require('../config.js');
 const awsLambdaTrigger = require('../triggers/aws_lambda.js');
 const eventInterface = require('../event.js');
 const lambdaRunner = require('../runners/aws_lambda.js');
 const { STEP_ID_NAME } = require('../consts.js');
 
+const FAILED_TO_SERIALIZE_MESSAGE = 'Unable to stringify response body as json';
+module.exports.FAILED_TO_SERIALIZE_MESSAGE = FAILED_TO_SERIALIZE_MESSAGE;
```
galbash, line 15: no other solution? do we really need to test it?

test/wrappers/test_lambda.js
```diff
@@ -7,8 +7,30 @@ const consts = require('../../src/consts.js');
...
+    return null;
+}
+
+function getReturnValue(addEventStub) {
+    const runnerMetadata = getRunner(addEventStub).getResource().getMetadataMap();
+    return runnerMetadata.get('return_value');
+}
 
 describe('lambdaWrapper tests', () => {
+    before(() => { config.setConfig({ metadataOnly: false }); });
```
galbash: please put it in the before each (i know currently its not needed but if someone adds a test that changes the config we will be fucked)
on another thought - please mock getConfig - so the config we get is local to the test and won't be affected by other tests

galbash (dismissed):
Looks really good, small comments

src/config.js
ronyis: This will break a lot of code, and especially tests because they are used to replace the config object.
This change is not related to this feature, I just had to add this to fix a problem with tests.

src/wrappers/lambda.js
ronyis: I think it's better to test it...

test/wrappers/test_lambda.js
ronyis: This whole `config` usage is very bad. What you suggest is just a small patch so I think I should just leave it as it.

test/runners/test_aws_batch.js
```diff
@@ -10,11 +10,14 @@ describe('AWS Batch createRunner tests', () => {
     beforeEach(() => {
         this.getStub = sinon.stub(axios, 'get').returns(Promise.resolve({ data: JSON.stringify({ region: 'test-reg' }) }));
         this.addExceptionStub = sinon.stub(tracer, 'addException');
-        config.setConfig({ metadataOnly: false });
+        this.getConfigStub = sinon.stub(config, 'getConfig').returns({
+            metadataOnly: false, // adding only members relevant to the test.
```
ronyis: This is not so good... you assume that this the only field in the metadata that is used

ronyis (commented):
What about `config` module tests? It still uses 'config.config'

test/runners/test_aws_batch.js
galbash: the tests are creating a stub with only the relevant data to the test. If more data will be relevant - we will add it

ranrib (approved):

PR #18, opened 2018-12-06
fix(lambda_wrapper): fixing async callback handling

galbash:


src/wrappers/lambda.js
```diff
@@ -135,12 +138,23 @@ function baseLambdaWrapper(
             );
 
             if (result instanceof Promise) {
+                let raisedError;
```
ranrib, line 141: `let raisedError = null`

ranrib (approved):

PR #20, opened 2018-12-07
fix(patcher.js): rollup does not build dynamic imports

ranrib:
rollup does not compile dynamic imports (such as tryRequire) in the build.
It means that events/pg was never instrumented.

### DIFF
--- a/src/events/pg.js
+++ b/src/events/pg.js
@@ -4,8 +4,8 @@ const tryRequire = require('try-require');
 /* eslint-disable import/no-unresolved, import/no-extraneous-dependencies */
 const pg = tryRequire('pg');
 const Pool = tryRequire('pg-pool');
-const sqlParser = tryRequire('node-sqlparser');
 /* eslint-enable import/no-unresolved, import/no-extraneous-dependencies */
+const sqlParser = require('node-sqlparser');
 const utils = require('../utils.js');
 const tracer = require('../tracer.js');
 const serverlessEvent = require('../proto/event_pb.js');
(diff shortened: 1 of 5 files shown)


src/events/pg.js
```diff
@@ -1,9 +1,10 @@
 const uuid4 = require('uuid4');
 const shimmer = require('shimmer');
+const tryRequire = require('try-require');
 /* eslint-disable import/no-unresolved, import/no-extraneous-dependencies */
-const pg = require('pg');
-const Pool = require('pg-pool');
-const sqlParser = require('node-sqlparser');
+const pg = tryRequire('pg');
+const Pool = tryRequire('pg-pool');
+const sqlParser = tryRequire('node-sqlparser');
```
galbash: This should be a regular require

galbash (commented):
1 small comment

src/events/pg.js
```diff
@@ -1,10 +1,11 @@
...
 const shimmer = require('shimmer');
+const tryRequire = require('try-require');
 /* eslint-disable import/no-unresolved, import/no-extraneous-dependencies */
-const pg = require('pg');
-const Pool = require('pg-pool');
-const sqlParser = require('node-sqlparser');
+const pg = tryRequire('pg');
+const Pool = tryRequire('pg-pool');
 /* eslint-enable import/no-unresolved, import/no-extraneous-dependencies */
+const sqlParser = require('node-sqlparser');
```
galbash, line 4: Remove the eslint line above

ranrib: The eslint line above is `enable` (comes after the `disable`)

ranrib: 👍

galbash (approved):

PR #27, opened 2018-12-18
feat(labels): support custom labels in traces

razlo:
Adding support for custom labels in users' traces.

### DIFF
--- a/examples/label_example.js
+++ b/examples/label_example.js
@@ -6,7 +6,7 @@ epsagon.init({
     metadataOnly: false,
 });
 
-module.exports.test = epsagon.lambdaWrapper(() => {
+module.exports.test = epsagon.lambdaWrapper(async () => {
     epsagon.label('myFirstLabel', 'customValue1');
     epsagon.label('mySecondLabel', 'customValue2');
 
--- a/examples/label_example.js
+++ b/examples/label_example.js
@@ -6,10 +6,9 @@ epsagon.init({
     metadataOnly: false,
 });
 
-module.exports.test = epsagon.lambdaWrapper((event, context, callback) => { // eslint-disable-line no-unused-vars
+module.exports.test = epsagon.lambdaWrapper(() => {
     epsagon.label('myFirstLabel', 'customValue1');
     epsagon.label('mySecondLabel', 'customValue2');
 
-    // eslint-disable-next-line no-console
-    console.log('hello world from node function');
+    return 'success';
 });
--- a/examples/label_example.js
+++ b/examples/label_example.js
@@ -1,0 +1,15 @@
+const epsagon = require('epsagon');
+
+epsagon.init({
+    token: 'my-secret-token',
+    appName: 'my-app-name',
+    metadataOnly: false,
+});
+
+module.exports.test = epsagon.lambdaWrapper((event, context, callback) => { // eslint-disable-line no-unused-vars
+    epsagon.label('myFirstLabel', 'customValue1');
+    epsagon.label('mySecondLabel', 'customValue2');
+
+    // eslint-disable-next-line no-console
+    console.log('hello world from node function');
+});
--- a/src/event.js
+++ b/src/event.js
@@ -69,3 +69,22 @@ module.exports.addObjectToMetadata = function addObjectToMetadata(
     }
     event.getResource().getMetadataMap().set(key, JSON.stringify(objectToAdd));
 };
+
+/**
+ * Adds items from a map to a resource Metadata
+ * @param {proto.event_pb.Event} event The event to add the items to
+ * @param {string} key key for the added label
+ * @param {string} value value for the added label
+ */
+module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, value) {
+    const currLabels = event.getResource().getMetadataMap().get('labels');
+    let labels = null;
+    if (currLabels !== undefined) {
+        labels = JSON.parse(currLabels);
+        labels[key] = value;
+    } else {
+        labels = { [key]: value };
+    }
+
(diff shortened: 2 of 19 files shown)


src/event.js
```diff
@@ -69,3 +69,22 @@ module.exports.addObjectToMetadata = function addObjectToMetadata(
...
 };
+
+/**
+ * Adds items from a map to a resource Metadata
+ * @param {proto.event_pb.Event} event The event to add the items to
+ * @param {string} key key for the added label
+ * @param {string} value value for the added label
+ */
+module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, value) {
+    const currLabels = event.getResource().getMetadataMap().get('labels');
```
galbash, line 81: you can use `const currLabels = bla_bla || {};`
only if you want

src/event.js
```diff
@@ -69,3 +69,22 @@ module.exports.addObjectToMetadata = function addObjectToMetadata(
     }
     event.getResource().getMetadataMap().set(key, JSON.stringify(objectToAdd));
 };
+
+/**
+ * Adds items from a map to a resource Metadata
```
galbash: this is not true?

src/tracer.js
```diff
@@ -23,6 +23,12 @@ module.exports.tracer = new trace.Trace([
     `node ${process.versions.node}`,
 ]);
 
+/**
+ * Custom traces that will be added to the runner of the current trace.
```
galbash: custom labels?

src/tracer.js
```diff
@@ -88,13 +94,19 @@ module.exports.initTrace = function initTrace(
 /**
  * Restarts the tracer. Has to be called after a trace has been sent to reset the tracer
  * and start collecting a new trace
+ * @param {object} runner The current of the current trace
```
galbash: the current of the current of the current of the current trace?

src/tracer.js
```diff
@@ -207,3 +219,19 @@ module.exports.sendTraceSync = function sendTraceSync() {
...
+ * @param {string} value value for the added label
+ */
+module.exports.label = function addLabel(key, value) {
+    if (typeof key !== 'string' || typeof value !== 'string') {
+        return;
+    }
+
+    if (key.length + value.length + currentLabelsLen <= consts.MAX_LABEL_SIZE) {
+        eventInterface.addLabelToMetadata(currRunner, key, value);
+        currentLabelsLen += key.length + value.length;
```
galbash: why do you need currentLabelsLen?

src/wrappers/lambda.js
```diff
@@ -33,12 +33,12 @@ function baseLambdaWrapper(
 ) {
     // eslint-disable-next-line consistent-return
     return (originalEvent, originalContext, originalCallback) => {
-        tracer.restart();
         let runner;
         let callbackCalled = false;
 
         try {
             runner = lambdaRunner.createRunner(originalContext, runnerResourceType);
```
galbash, line 41: I think this will be problemtic since if we fail in runner creation we will send the previous trace. Maybe we have to split it to restart and addRunner

galbash (changes requested):
Take a look at the comments and tell me what you think :)

src/tracer.js
razlo: lol

src/tracer.js
razlo: to avoid dynamically calculating the size of the map each time
is there any nice JS way to do it? (couldn't find something)

examples/label_example.js
```diff
@@ -0,0 +1,23 @@
+const epsagon = require('../src/index.js');
```
galbash: maybe change this to require('epsagon'), doesn't really matter that it doesn't work cause users will copy paste it.
Maybe have a look at other repos examples and see what is the standard?

examples/label_example.js
```diff
@@ -0,0 +1,23 @@
...
+    epsagon.label('myFirstLabel', 'customValue1');
+    epsagon.label('mySecondLabel', 'customValue2');
+
+    // eslint-disable-next-line no-console
+    console.log('hello world from node function');
+}
+
+const testFunction = epsagon.nodeWrapper(test);
+
+testFunction(() => {});
```
galbash: no need

examples/label_example.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+function test(callback) { // eslint-disable-line no-unused-vars
+    epsagon.label('myFirstLabel', 'customValue1');
+    epsagon.label('mySecondLabel', 'customValue2');
+
+    // eslint-disable-next-line no-console
+    console.log('hello world from node function');
+}
+
+const testFunction = epsagon.nodeWrapper(test);
```
galbash: use a lambda wrapper, the nodeWrapper is not documened.

galbash: also export this

src/tracer.js
galbash: we talked about it, fix it <3

src/tracer.js
```diff
@@ -85,9 +91,26 @@ module.exports.initTrace = function initTrace(
...
+ * Adds a runner to the current trace.
+ * @param {object} runner The runner of the current trace
+ * @param {Promise} runnerPromise A promise that resolves when the event handling is Done,
+ *      if required.
+ */
+module.exports.addRunner = function addRunner(runner, runnerPromise) {
+    const { tracer } = module.exports;
+    tracer.addEvent(runner, runnerPromise);
+    currRunner = runner;
+    currentLabelsLen = 0;
```
galbash: move this to restart (actually just remove it)

galbash (changes requested):

galbash (approved):

PR #48, opened 2019-01-30
feat(mysql): adding support for mysql and mysql2, improving pg tracing

ronyis:


src/events/sql.js
```diff
@@ -0,0 +1,111 @@
...
+            sqlObj = sqlParser.parse(queryString);
+        } catch (error) {
+            sqlObj.type = 'SQL-Command';
+        }
+
+        const { type, table } = sqlObj;
+
+        const { database, host } = config;
+
+        let resourceType = 'sql';
```
ranrib, line 49: `= 'database'`

ronyis: 'database' is too generic, another resource could have the same type and name.

ranrib (approved):

PR #52, opened 2019-02-08
fix(sql.js): fix sql callback extraction.

ranrib:
Fixing:
1. just-extend vuln.
2. callback extraction in mysql.
3. parsing query in sequelize.
4. printing full trace.

### DIFF
--- a/src/events/mysql.js
+++ b/src/events/mysql.js
@@ -17,14 +17,14 @@ function mysqlQueryWrapper(wrappedFunction) {
         let params;
         if (typeof sql !== 'string') {
             queryString = sql.sql;
-            params = sql.values;
-            callback = (
-                sql.onResult ||
-                (typeof arg2 === 'function' && arg2) ||
-                (typeof arg1 === 'function' && arg1)
-            );
         } else {
             queryString = sql;
+        }
+
+        if (sql.onResult) {
+            params = sql.values;
+            callback = sql.onResult;
+        } else {
             ({ params, callback } = sqlWrapper.parseQueryArgs(arg1, arg2));
         }
 
--- a/src/events/mysql.js
+++ b/src/events/mysql.js
@@ -15,14 +15,17 @@ function mysqlQueryWrapper(wrappedFunction) {
         let queryString;
         let callback;
         let params;
-        // This function can be called with either a string or Query object
-        if (typeof sql === 'string') {
-            queryString = sql;
-            ({ params, callback } = sqlWrapper.parseQueryArgs(arg1, arg2));
-        } else {
+        if (typeof sql !== 'string') {
             queryString = sql.sql;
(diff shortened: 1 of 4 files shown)


src/events/mysql.js
```diff
@@ -15,14 +15,17 @@ function mysqlQueryWrapper(wrappedFunction) {
         let queryString;
         let callback;
         let params;
-        // This function can be called with either a string or Query object
-        if (typeof sql === 'string') {
-            queryString = sql;
-            ({ params, callback } = sqlWrapper.parseQueryArgs(arg1, arg2));
-        } else {
+        if (typeof sql !== 'string') {
```
ronyis, line 18: I think there is a little issue with that:
We have seen that in 'sequalize', the object is not `Commands.Query` but still has an `sql` field containing the query.
Although, it may not have the `values` field set, and both `values` and `callback` can be given as arguments (see `mysql2.connection.createQuery`). But you only take `callback` argument in this case.
Seems to me that the correct thing to do is to check if `onResult` is set, and decide by it wether to take both arguments from the `sql` object or from this function's arguments/

src/events/sql.js
```diff
@@ -37,7 +37,9 @@ module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callbac
     try {
         let sqlObj = {};
         try {
-            sqlObj = sqlParser.parse(queryString);
+            // Sanitizing query.
+            const queryStringSan = queryString.split('`').join('').replace(';', '');
```
ronyis: Why not just `queryString.replace('`', '')`?
And why do you want to remove ';'? It used to concatenate several queries

src/tracer.js
```diff
@@ -177,7 +176,7 @@ function sendCurrentTrace(traceSender) {
  *  */
 function postTrace(traceObject) {
     utils.debugLog(`Posting trace to ${config.getConfig().traceCollectorURL}`);
-    utils.debugLog(`trace: ${util.inspect(traceObject)}`);
+    utils.debugLog(`trace: ${JSON.stringify(traceObject, null, 2)}`);
```
ronyis, line 179: Avoid printing the whole JSON to log?

ronyis (changes requested):

src/events/sql.js
ranrib: replace in js just replaces the first instance :(

src/tracer.js
ranrib: It's annoying - if we're in debug mode I prefer to see the full payload

src/events/mysql.js
ranrib: But sometimes `sql.onResult` is undefined, and `sql.sql` is defined.

ronyis (approved):

PR #53, opened 2019-02-10
feat(wrapper): sending trace before function timeout is reached

liadgo:


package.json
```diff
@@ -57,6 +57,7 @@
     "eslint-plugin-mocha": "^4.11.0",
     "eslint-plugin-react": "^7.11.0",
     "husky": "^1.1.0",
+    "lolex": "^3.0.0",
```
galbash, line 60: ?

package.json
```diff
@@ -81,6 +82,7 @@
     "json.sortify": "^2.2.2",
     "md5": "^2.2.1",
     "node-sqlparser": "^1.0.4",
+    "nvm": "0.0.4",
```
galbash: this shouldn't be a dependency

src/event.js
```diff
@@ -28,6 +28,14 @@ module.exports.setException = function setException(event, error) {
     }
 };
 
+/**
+ * Add timeout indication to a given event
+ * @param {proto.event_pb.Event} event The event the timeout is set on
+ */
+module.exports.setTimeout = function setTimeout(event) {
```
galbash: maybe markAsTimeout? (setTimeout is kinda reserved in node)

src/proto/error_code_pb.js
```diff
@@ -18,7 +18,8 @@ goog.exportSymbol('proto.ErrorCode', null, global);
 proto.ErrorCode = {
   OK: 0,
   ERROR: 1,
-  EXCEPTION: 2
+  EXCEPTION: 2,
+  TIMEOUT: 3
```
galbash, line 22: did you change epsagon-proto as well?

src/wrappers/lambda.js
```diff
@@ -106,6 +116,9 @@ function baseLambdaWrapper(
             // eslint-disable-next-line no-underscore-dangle
             process._events.beforeExit = originalBeforeExit;
 
+            // Mark trace as sent.
+            tracesSent = true;
```
galbash, line 117: whats the difference between this and callbackCalled?

src/wrappers/lambda.js
```diff
@@ -180,6 +193,11 @@ function baseLambdaWrapper(
         });
 
         try {
+            timeout = setTimeout(() => {
```
galbash: timeoutHandler

src/wrappers/lambda.js
galbash: I think I understand the difference now, but it is confusing. Maybe unite `callbackCalled` and `traceSent` to `traceSent` or somea better name or something?

galbash (changes requested):
take a look at the comments and tell me what you think

package.json
liadgo: i'm using it in order to manipulate the test's clock.

src/event.js
liadgo: K

src/proto/error_code_pb.js
liadgo: :)

src/wrappers/lambda.js
liadgo: Kkk

src/wrappers/lambda.js
```diff
@@ -83,6 +88,11 @@ function baseLambdaWrapper(
         };
 
         const handleUserExecutionDone = (error, result, sendSync) => {
+            clearTimeout(timeout);
+
+            if (tracesSent) {
```
galbash: if with ||

galbash (approved):
just making sure you checked it a bit with sync/ async / callback/ context.fail/ context.succeed etc to make sure we didn't break anything right?

PR #58, opened 2019-02-21
feat(tracer): add custom errors

ranrib:
Letting users send custom errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,20 @@ function handler(event, context, callback) {
 }
 ```
 
+## Custom errors
+
+You can set a trace as an error (although handled correctly) by sending an Error object:
+```node
+function handler(event, context, callback) {
+    try {
+        // something bad happens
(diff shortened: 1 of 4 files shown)


henperez (approved):

henperez (approved):

tavolate:
It will be useful if setError will set the transaction to "Error" so that the error counter is incremented.

This feature will be more useful for lambda triggered by API events. In fact, when an "unexpected" exception occurs, we send back to the user a human message instead of a raw error message. To accomplish that, we need to catch the exception and use setError to notify to Epsagon the error. But in this way, the transaction is set as "Success" instead of "Error"

ranrib:
Hi @tavolate, this is already setting the event to error code = 2 (exception), so it can be filtered in the dashboard. Can you describe what's missing? We can move to the intercom in the product if you would like

PR #67, opened 2019-03-06
feat(acceptance): initial sanity acceptance tests

enoodle:
Adding sanity acceptance tests.


galbash:
matrix of things to verify:

1. runtime
  a. 6, 8
2. function type
  b. sync / async (8)
3. handler
  a. no parameters
  b. event
  c. event  + context
  d. event + context + callback
4. return type (sucess / error)
 a. call callback
 b.  context.succeed(), context.fail(), context.done()
 c. (only in node 8) return promise (return value async function / actually returning a promise)
5. mix return types

test/acceptance/lambda-handlers/bad_events.json
```diff
@@ -0,0 +1,3 @@
+{
```
galbash: I think you can delete this file

test/acceptance/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest
```
galbash: You can delete this file

galbash (dismissed):
Looks good

galbash (approved):

PR #69, opened 2019-03-21
feat(acceptance): adding acceptance matrix

enoodle:


test/acceptance/acceptance.js
```diff
@@ -39,6 +71,32 @@ async function basicTest(lambdaName, input) {
...
+/**
+ * basicNoReturn
+ *
+ * @param {string} lambdaName
+ * @param {object} input
+ * @returns {undefined}
+ */
+async function basicNoReturn(lambdaName, input) {
+    const response = await invoke(lambdaName, input);
+    expect(response.StatusCode).to.equal(200);
```
galbash, line 83: Maybe assert the payload is 'null'?

test/acceptance/acceptance.js
```diff
@@ -55,7 +113,72 @@ describe('Lambda Wrapper', () => {
...
+        });
+    });
+
+    const syncOptions = ['sync'];
+    const paramOptions = ['no', '1', '2', '3'];
+    const returnOptions = [
+        'simple_return', 'no_return',
+        'succeed', 'fail', 'done',
+        'callback', 'callback_error', 'callback_then_fail'];
+    if (RUNTIME === 'nodejs8.10') {
```
galbash, line 127: Maybe on node6.10 push promise-void or something whereyou return a promise but expect the result to be null

galbash (dismissed):
Check out the comments

test/acceptance/acceptance.js
enoodle: then it is like no returning anything which we already check..

test/acceptance/acceptance.js
enoodle: :+1: will add

galbash (approved):

PR #71, opened 2019-03-26
fix(tracer.js): send traces in session

ranrib:
Use the same session

### DIFF
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -45,7 +45,6 @@ const session = axios.create({
     httpAgent: new http.Agent({ keepAlive: true }),
     httpsAgent: new https.Agent({ keepAlive: true }),
 });
-module.exports.sessionPost = session.post;
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/tracer.js
```diff
@@ -31,9 +33,19 @@ let currRunner = null;
...
+
+/**
+ * Session for the post requests to the collector
+ */
+const session = axios.create({
+    timeout: sendTimeoutMilliseconds,
+    httpAgent: new http.Agent({ keepAlive: true }),
+    httpsAgent: new https.Agent({ keepAlive: true }),
+});
+module.exports.sessionPost = session.post;
```
galbash: why are you exporting this? if this is for the UT you should mock the result of axios.create instead

src/tracer.js
```diff
@@ -177,10 +189,9 @@ function sendCurrentTrace(traceSender) {
 function postTrace(traceObject) {
     utils.debugLog(`Posting trace to ${config.getConfig().traceCollectorURL}`);
     utils.debugLog(`trace: ${JSON.stringify(traceObject, null, 2)}`);
-    return axios.post(
+    return module.exports.sessionPost(
```
galbash: is there no way the connection will be closed? (a keep alive might not be sent because the container did not get enough runtime?) maybe we need to check and re-open if needed?

galbash (commented):
few small questions

src/tracer.js
ranrib: tested the following scenarios:
1. HTTP server (i.e. TC) sends a Connection: closed header.
2. server running socket.close event (fin).

In both case it works OK

ranrib:
@galbash tested on the same vm instance (same log stream) in 5 and 20 minutes interval - all worked.

galbash (approved):

PR #74, opened 2019-04-16
feat(express.js): add support for express

ranrib:
Letting users instrument express apps on non Lambda environments

### DIFF
--- a/examples/express_example.js
+++ b/examples/express_example.js
@@ -11,4 +11,10 @@ const app = express()
 
 app.get('/', (req, res) => res.send('Hello World!'))
 
+app.get('/label_example', (req, res) => {
+    // Example label usage
+    req.epsagon.label('myFirstLabel', 'customValue1');
+    res.send('Hello World!'))
+}
+
 app.listen(3000)
--- a/examples/express_example.js
+++ b/examples/express_example.js
@@ -1,0 +1,14 @@
+const express = require('express');
+const epsagon = require('epsagon');
+
+epsagon.init({
+    token: 'my-secret-token',
+    appName: 'my-app-name',
+    metadataOnly: false,
+});
+
+const app = express()
+
+app.get('/', (req, res) => res.send('Hello World!'))
+
+app.listen(3000)
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   },
   "dependencies": {
     "axios": "^0.18.0",
+    "cls-hooked": "^4.2.2",
     "google-protobuf": "^3.5.0",
     "json.sortify": "^2.2.2",
     "md5": "^2.2.1",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint-plugin-jsx-a11y": "^6.1.1",
     "eslint-plugin-mocha": "^4.11.0",
     "eslint-plugin-react": "^7.11.0",
+    "express": "^4.16.4",
     "husky": "^1.1.0",
     "lolex": "^3.0.0",
     "mocha": "^5.0.4",
@@ -71,7 +72,7 @@
     "rollup-plugin-eslint": "^5.0.0",
     "rollup-plugin-json": "^3.1.0",
     "rollup-plugin-terser": "^3.0.0",
-    "semantic-release": "^15.9.16",
+    "semantic-release": "^15.13.12",
     "sinon": "^4.3.0",
     "uglify-es": "^3.3.9"
   },
--- a/src/events/express.js
+++ b/src/events/express.js
@@ -1,0 +1,71 @@
+/* eslint-disable prefer-rest-params */
+/**
+ * @fileoverview Handlers for Express instrumentation
+ */
+
+const shimmer = require('shimmer');
+const tryRequire = require('../try_require.js');
+const utils = require('../utils.js');
+const tracer = require('../tracer.js');
+const expressRunner = require('../runners/express.js');
+
+const express = tryRequire('express');
+
+
+/**
+ * Wraps the Express module request function with tracing
+ * @param {Function} wrappedFunction The Express's request module
+ * @returns {Function} The wrapped function
+ */
+function expressWrapper(wrappedFunction) {
+    return function internalExpressWrapper() {
+        const req = arguments[0];
+        if (!req.next) {
+            return wrappedFunction.apply(this, arguments);
+        }
+        tracer.restart();
+        const res = arguments[1];
+        let expressEvent;
+        const startTime = Date.now();
+        const result = wrappedFunction.apply(this, arguments);
+
+        if (!req.route) {
+            return result;
+        }
+
+        try {
+            expressEvent = expressRunner.createRunner(req, startTime);
+        } catch (err) {
+            utils.debugLog(err);
+        }
+
+        res.on('finish', () => {
+            try {
+                expressRunner.finishRunner(expressEvent, this, startTime);
+            } catch (err) {
+                tracer.addException(err);
+            }
+        });
+
+        return result;
+    };
+}
+
+module.exports = {
+    /**
+     * Initializes the Express tracer
+     */
+    init() {
+        if (express && express.Router) {
+            shimmer.wrap(express.Router, 'handle', expressWrapper);
+        }
+    },
+    /**
+     * Un-patches the Express tracer
+     */
+    unpatch() {
+        if (express && express.Router) {
+            shimmer.unwrap(express.Router, 'handle');
+        }
+    },
+};
--- a/src/patcher.js
+++ b/src/patcher.js
@@ -22,7 +22,7 @@ function patch(patcher) {
         patcher.init();
     } catch (error) {
         if ((process.env.EPSAGON_DEBUG || '').toUpperCase() === 'TRUE') {
-            console.log(error); // eslint-disable-line no-console
+            utils.debugLog(error);
         }
     }
 }
--- a/src/patcher.js
+++ b/src/patcher.js
@@ -9,6 +9,7 @@ const pgPatcher = require('./events/pg.js');
 const mysqlPatcher = require('./events/mysql.js');
 const redisPatcher = require('./events/redis.js');
 const mongoPatcher = require('./events/mongodb.js');
+const expressPatcher = require('./events/express.js');
 
 
 if (!config.getConfig().isEpsagonPatchDisabled) {
@@ -19,6 +20,7 @@ if (!config.getConfig().isEpsagonPatchDisabled) {
         mysqlPatcher,
         redisPatcher,
         mongoPatcher,
+        expressPatcher,
     ].forEach((patcher) => {
         try {
             patcher.init();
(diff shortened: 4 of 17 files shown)


package.json
```diff
@@ -71,7 +72,7 @@
     "rollup-plugin-eslint": "^5.0.0",
     "rollup-plugin-json": "^3.1.0",
     "rollup-plugin-terser": "^3.0.0",
-    "semantic-release": "^15.9.16",
+    "semantic-release": "^15.13.12",
```
ranrib, line 75: This is due to npm audit command

src/events/express.js
```diff
@@ -0,0 +1,71 @@
...
+
+const express = tryRequire('express');
+
+
+/**
+ * Wraps the Express module request function with tracing
+ * @param {Function} wrappedFunction The Express's request module
+ * @returns {Function} The wrapped function
+ */
+function expressWrapper(wrappedFunction) {
```
ronyis: Looks like it should be in the `wrappers` dir

src/runners/express.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {Response} res response data
+ * @param {Int} startTime Runner start time
+ */
+function finishRunner(expressEvent, res, startTime) {
+    eventInterface.addToMetadata(expressEvent, {
+        status_code: this.statusCode,
+    }, {
+        response_headers: this.headers,
+    });
+    if (this.statusCode >= 300) {
```
ronyis: It should be ">=500"
3xx - Redirection
4xx - Client error

src/runners/express.js
```diff
@@ -0,0 +1,67 @@
...
+ * Terminates the running Express (runner)
+ * @param {Object} expressEvent runner's express event
+ * @param {Response} res response data
+ * @param {Int} startTime Runner start time
+ */
+function finishRunner(expressEvent, res, startTime) {
+    eventInterface.addToMetadata(expressEvent, {
+        status_code: this.statusCode,
+    }, {
+        response_headers: this.headers,
```
ronyis: Better to shorten if it's too big (maybe add a function to `utils`)

src/runners/express.js
```diff
@@ -0,0 +1,67 @@
...
+    });
+    if (this.statusCode >= 300) {
+        expressEvent.setErrorCode(errorCode.ErrorCode.EXCEPTION);
+    }
+    expressEvent.setDuration(utils.createDurationTimestamp(startTime));
+    tracer.addRunner(expressEvent);
+    tracer.sendTrace(() => {});
```
ronyis: It makes more sense to do that in the wrapper

ronyis: I think the express can handle more than a single trace at a time.
Seems we need a different `tracer` like object for each one.
Talk with @galbash

ronyis: We should consider adding a trigger for this case, it could make more sense when merging this trace to the calling one.
Also, now the duration saved is only till the response is sent, but it doesn't mean that the handler finished (there could be pending events).
I'm not sure about the right solution but we should think about it.

src/events/express.js
```diff
@@ -0,0 +1,71 @@
...
+            expressEvent = expressRunner.createRunner(req, startTime);
+        } catch (err) {
+            utils.debugLog(err);
+        }
+
+        res.on('finish', () => {
+            try {
+                expressRunner.finishRunner(expressEvent, this, startTime);
+            } catch (err) {
+                tracer.addException(err);
```
ronyis: Currently, you call `addException` even when sending the trace failed.
In this case, you should print a debug log.

examples/express_example.js
```diff
@@ -0,0 +1,14 @@
...
+
+epsagon.init({
+    token: 'my-secret-token',
+    appName: 'my-app-name',
+    metadataOnly: false,
+});
+
+const app = express()
+
+app.get('/', (req, res) => res.send('Hello World!'))
```
ronyis, line 12: It's not a good idea to add the wrapper implicitly. Sometimes the user will have an express application with wrappers in other places. And it makes this library much complicated (needing to make sure a wrapper is not added twice)

ronyis (changes requested):
Many problems

src/events/express.js
ranrib: It's not similar to any wrapper in the wrappers dir, since it's not manual. the name wrapper that written here is because it wraps a function (this name is similar to all other `events/*` files)

src/runners/express.js
ranrib: 👍

src/runners/express.js
ranrib: 👍

src/events/express.js
ranrib: Fixed with before

src/patcher.js
```diff
@@ -3,13 +3,29 @@
...
+
+/**
+ * Patches a module
+ * @param {Object} patcher module
+ */
+function patch(patcher) {
+    try {
+        patcher.init();
+    } catch (error) {
+        if ((process.env.EPSAGON_DEBUG || '').toUpperCase() === 'TRUE') {
```
ronyis, line 24: Any reason for not using `utils.debugLog`?

src/patcher.js
```diff
@@ -20,15 +36,12 @@ if (!config.getConfig().isEpsagonPatchDisabled) {
...
-            if ((process.env.EPSAGON_DEBUG || '').toUpperCase() === 'TRUE') {
-                // eslint-disable-next-line no-console
-                console.log(error);
-            }
-        }
-    });
+    ].forEach(patch);
+
+    // Conditional patching that depends on the environment
+    if (!utils.isLambdaEnv()) {
```
ronyis, line 42: That's better.. But still not too future-compatible

src/events/express.js
ronyis: It's a wrapper, like a wrapper for lambda. It just happened to work with a patch. It's definitely not like other events in this dir.

src/runners/express.js
ronyis: What about that?
Your test doesn't seem to make sure it can't happen (it probably can).

ronyis: What about that?
Your test doesn't seem to make sure it can't happen (it probably can).

ranrib:
@galbash @ronyis I've added a trace context. Would love to get your review please

src/tracer.js
```diff
@@ -11,26 +11,38 @@ const utils = require('./utils.js');
...
+module.exports.createTracer = function createTracer() {
+    const tracerObj = new trace.Trace([
+        '',
+        '',
+        [],
+        [],
+        consts.VERSION,
+        `node ${process.versions.node}`,
+    ]);
+    tracerObj.currRunner = null;
```
ronyis: Don't change the proto object, wrap it in another object

src/tracer.js
```diff
@@ -11,26 +11,38 @@ const utils = require('./utils.js');
...
 
-let currRunner = null;
 
 /**
- * The requests promises pending to resolve. All must be resolved before sending the trace.
- * A Map containing (event, promise) pairs.
+ * The tracer singleton, used to manage the trace and send it at the end of the function invocation
  */
-const pendingEvents = new Map();
+module.exports.tracer = module.exports.createTracer();
```
ronyis: Consider creating the singleton only when needed

src/trace_context.js
```diff
@@ -0,0 +1,36 @@
...
+const cls = require('cls-hooked');
+
+const namespaceId = 'epsagon-context';
+const namespace = cls.createNamespace(namespaceId);
+
+
+/**
+ * Creates an active context for tracer
+ * @param {Trace} tracer object
+ * @param {Request} next The Express's next function
```
ronyis: It doesn't have to be Express

src/trace_context.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Function} The runner event
+ */
+function activateContext(tracer, next) {
```
ronyis: Call it `RunInContext` or something

src/wrappers/express.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Express requests middleware
+ * @param {Request} req The Express's request data
+ * @param {Response} res The Express's response data
+ * @param {Function} next express function
+ */
+function expressMiddleware(req, res, next) {
+    // Initialize tracer
+    const tracerObj = tracer.createTracer();
+    tracer.restart(tracerObj);
```
ronyis, line 26: Seems redundant

src/runners/express.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * @fileoverview Runner for Express application
+ */
+const uuid4 = require('uuid4');
+const utils = require('../utils.js');
+const serverlessEvent = require('../proto/event_pb.js');
```
ronyis: That's not a good name, just call it `event`

ronyis (changes requested):

src/wrappers/express.js
ranrib: It sets the appname and the token

src/runners/express.js
```diff
@@ -0,0 +1,65 @@
...
+        `express-${uuid4()}`,
+        utils.createTimestampFromTime(startTime),
+        null,
+        'runner',
+        0,
+        errorCode.ErrorCode.OK,
+    ]);
+
+    const resource = new serverlessEvent.Resource([
+        req.path,
```
ronyis: Are you sure that it's a template and not an actual path? (which can get many different values).
What about connecting this event with the request? Maybe we need to add a strong ID to the response.
And the resource name in requests is the domain. If we wish to update it to the path, we need to do that in the backend.

src/runners/express.js
```diff
@@ -0,0 +1,65 @@
...
+ * Terminates the running Express (runner)
+ * @param {Object} expressEvent runner's express event
+ * @param {Response} res response data
+ * @param {Request} req The Express's request data
+ * @param {Int} startTime Runner start time
+ */
+function finishRunner(expressEvent, res, req, startTime) {
+    expressEvent.getResource().setName(req.route.path);
+    eventInterface.addToMetadata(expressEvent, {
+        url: `${req.protocol}://${req.hostname}${req.path}`,
```
ronyis, line 47: Seems here that the path is not a template...
And I think that URL encoding is missing (not sure how important it is)

src/tracer.js
ranrib: Added more docs around it.

ronyis (approved):

src/runners/express.js
ronyis: .

ronyis (changes requested):

src/trace_object.js
```diff
@@ -0,0 +1,36 @@
...
+ * The tracer singleton, used to manage the trace and send it at the end of the function invocation.
+ * In a Lambda environment we use this singleton, while in other environment we use the one from
+ * the context.
+ */
+module.exports.tracer = null;
+
+
+/**
+ * Creates a new Trace object
+ * @returns {Trace} new Trace
```
ronyis: Update

src/wrappers/express.js
ronyis: So the function name is wrong but ok..

ronyis (approved):
Some names are not so good, but looks ok

PR #78, opened 2019-04-24
Support express

ranrib:
fixing express support by adding `if (!tracerObj) return;` in 2 locations

### DIFF
--- a/src/trace_object.js
+++ b/src/trace_object.js
@@ -8,7 +8,7 @@ const tracerModule = require('./tracer.js');
  * In a Lambda environment we use this singleton, while in other environment we use the one from
  * the context.
  */
-module.exports.tracer = null;
+let tracer = null;
 
 
 /**
@@ -16,8 +16,8 @@ module.exports.tracer = null;
  * @returns {Object} tracer object
  */
 module.exports.get = () => {
-    if (!module.exports.tracer) {
-        module.exports.tracer = tracerModule.createTracer();
+    if (!tracer) {
+        tracer = tracerModule.createTracer();
     }
-    return module.exports.tracer;
+    return tracer;
 };
--- a/src/trace_object.js
+++ b/src/trace_object.js
@@ -1,5 +1,7 @@
-const trace = require('./proto/trace_pb.js');
-const consts = require('./consts.js');
+/**
+ * @fileoverview holds tracer singleton.
+ */
+const tracerModule = require('./tracer.js');
 
 /**
  * The tracer singleton, used to manage the trace and send it at the end of the function invocation.
@@ -10,27 +12,12 @@ module.exports.tracer = null;
 
 
 /**
- * Creates a new Trace object
- * @returns {Trace} new Trace
+ * The tracer singleton getter function
+ * @returns {Object} tracer object
  */
-module.exports.createTracer = function createTracer() {
-    const tracerObj = new trace.Trace([
-        '',
-        '',
-        [],
-        [],
-        consts.VERSION,
-        `node ${process.versions.node}`,
-    ]);
-    // The requests promises pending to resolve. All must be resolved before sending the trace.
-    // A Map containing (event, promise) pairs.
-    return {
-        trace: tracerObj,
-        currRunner: null,
-        pendingEvents: new Map(),
-    };
+module.exports.get = () => {
+    if (!module.exports.tracer) {
+        module.exports.tracer = tracerModule.createTracer();
+    }
+    return module.exports.tracer;
 };
-
-module.exports.get = () => module.exports.tracer;
-
-module.exports.init = () => module.exports.tracer = module.exports.createTracer();
--- a/src/trace_object.js
+++ b/src/trace_object.js
@@ -1,0 +1,36 @@
+const trace = require('./proto/trace_pb.js');
+const consts = require('./consts.js');
+
+/**
+ * The tracer singleton, used to manage the trace and send it at the end of the function invocation.
+ * In a Lambda environment we use this singleton, while in other environment we use the one from
+ * the context.
+ */
+module.exports.tracer = null;
+
+
+/**
+ * Creates a new Trace object
+ * @returns {Trace} new Trace
+ */
+module.exports.createTracer = function createTracer() {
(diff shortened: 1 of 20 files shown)


src/trace_object.js
```diff
@@ -0,0 +1,23 @@
...
+ * @fileoverview holds tracer singleton.
+ */
+const tracerModule = require('./tracer.js');
+
+/**
+ * The tracer singleton, used to manage the trace and send it at the end of the function invocation.
+ * In a Lambda environment we use this singleton, while in other environment we use the one from
+ * the context.
+ */
+module.exports.tracer = null;
```
galbash: don't export it (there is a function to get it)

src/tracer.js
```diff
@@ -12,25 +12,40 @@ const config = require('./config.js');
...
-]);
+module.exports.traceGetter = () => {};
 
-let currRunner = null;
+/**
+ * Returns the relevant tracer. If got one as a param, or from active context, or singleton.
+ * @param {Object} tracer Optional tracer
+ * @returns {Object} active tracer
+ */
+const getTracer = tracer => tracer || module.exports.traceGetter();
```
galbash: remove this. have the functions use traceGetter and not pass along the tracer, it doesn't make sense

galbash: (using two design patterns, one is redundant)

src/tracer.js
```diff
@@ -46,26 +61,29 @@ const session = axios.create({
...
 
-
 /**
  * Adds an event to the tracer
  * @param {proto.event_pb.Event} event The event to add
  * @param {Promise} [promise] A promise that resolves when the event handling is Done, if required.
+ * @param {Object} tracer Optional tracer
  */
-module.exports.addEvent = function addEvent(event, promise) {
+module.exports.addEvent = function addEvent(event, promise, tracer) {
```
galbash: all these functions should not get the tracer :)

src/utils.js
```diff
@@ -51,8 +51,17 @@ function debugLog(...args) {
     }
 }
 
+/**
+ * Detects if code is running on Lambda environment
+ * @return {Boolean} true if running on Lambda, otherwise false
+ */
+function isLambdaEnv() {
+    return !!process.env.LAMBDA_TASK_ROOT;
```
galbash, line 59: maybe find a better one? I think it might not exist on some custom runtimes? (or just validate it does)

galbash: also might make more sense to invoke the patcher in the wrapper initialization instead of epsagon initialization, this way every wrapper can patch what he needs

src/wrappers/batch.js
```diff
@@ -11,6 +12,7 @@ const batchRunner = require('../runners/aws_batch.js');
  * Epsagon's batch Job wrapper, sets tracing on a batch job
  */
 module.exports.wrapBatchJob = function wrapBatchJob() {
+    tracer.traceGetter = traceObject.get;
```
galbash: why not keep the name? What I would do is make traceGetter have the traceObject/traceContext object, which both implement a `get` interface. then to invoke the events / tracer would use `traceGetter.get()`

src/wrappers/express.js
```diff
@@ -0,0 +1,84 @@
...
+    try {
+        expressEvent = expressRunner.createRunner(req, startTime);
+        tracer.addRunner(expressEvent, undefined, tracerObj);
+    } catch (err) {
+        utils.debugLog(err);
+    }
+
+    // Inject trace functions
+    req.epsagon = {
+        label: tracer.label,
```
galbash: use object destructing (also in setError)

src/wrappers/batch.js
galbash: also i would have a traceNoop which does nothing, and that's what the tracer is default value is (initialized with it) so we won't have errors like the ones we had

galbash (changes requested):
some comments

src/wrappers/express.js
ranrib: Done

src/trace_object.js
ranrib: Done

src/tracer.js
ranrib: Done

src/tracer.js
ranrib: Done

src/tracer.js
```diff
@@ -46,24 +54,27 @@ const session = axios.create({
     httpsAgent: new https.Agent({ keepAlive: true }),
 });
 
-
 /**
  * Adds an event to the tracer
  * @param {proto.event_pb.Event} event The event to add
  * @param {Promise} [promise] A promise that resolves when the event handling is Done, if required.
+ * @param {Object} tracer Optional tracer
```
galbash: Remove doc (everywhere)

src/tracer.js
```diff
@@ -46,24 +54,27 @@ const session = axios.create({
...
  */
 module.exports.addEvent = function addEvent(event, promise) {
+    const tracerObj = module.exports.getTrace();
+    if (!tracerObj) return;
```
galbash: Use {}  we dont charge you by the character

src/wrappers/express.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Express requests middleware
+ * @param {Request} req The Express's request data
+ * @param {Response} res The Express's response data
+ * @param {Function} next express function
+ */
+function expressMiddleware(req, res, next) {
+    // Initialize tracer
+    const tracerObj = tracer.createTracer();
+    tracer.getTrace = traceContext.get;
```
galbash: You should set the get method in initialization, not execution

galbash (changes requested):
Few small style and doc commsnts, one excution to initialization issue

src/tracer.js
ranrib: Done

src/wrappers/express.js
ranrib: Done

src/tracer.js
ranrib: 😄

galbash (approved):
LGTM

PR #95, opened 2019-05-26
fix(pg): adding promise support

galbash:
Adding support for the promise usage of the pg module. Prior to this PR anyone trying to use the promise version of the API would not get the promise he was expecting


ranrib (dismissed):

ronyis:
@ranrib  Don't merge yet, I should talk about it with Gal

ranrib (approved):

PR #99, opened 2019-06-05
User defined blacklist for HTTP filtering

liadgo:


src/events/http.js
```diff
@@ -36,6 +36,25 @@ function isBlacklistURL(url) {
...
 }
 
+/**
+ * Checks if a URL is in the user-defined blacklist.
+ * @param {string} url The URL to check
+ * @returns {boolean} True if it is in the user-defined blacklist, False otherwise.
+ */
+function isURLIgnoredByUser(url) {
+    let found = false;
+    config.getConfig().urlPatternsToIgnore.forEach(
```
ranrib: `return config.getConfig().urlPatternsToIgnore.some(pattern => url.includes(pattern))`

src/config.js
```diff
@@ -13,6 +13,7 @@ const config = {
     useSSL: (process.env.EPSAGON_SSL || 'TRUE').toUpperCase() === 'TRUE',
     traceCollectorURL: consts.TRACE_COLLECTOR_URL,
     isEpsagonDisabled: (process.env.DISABLE_EPSAGON || '').toUpperCase() === 'TRUE',
+    urlPatternsToIgnore: [],
```
ranrib, line 16: Let add here the ability to send params through env as well (I added this to Python as well).
the name should be `EPSAGON_URLS_TO_IGNORE` and it should include comma-separated patterns like
`book.com,la.co.il`

ranrib (changes requested):
Good! few comments

ranrib (approved):

PR #102, opened 2019-06-08
Create index.d.ts

### LINKED ISSUE #100: TypeScript type definitions
This is something I'm using internally in our TypeScript project. It'd be great if epsagon included `index.d.ts` inside of itself. 

```ts
// epsagon.d.ts
declare module 'epsagon' {
  import 'aws-lambda'

  export function init(options: {
    token: string
    appName: string
    metadataOnly?: boolean
  }): void
  export function label(key: string, value: string): void
  export function setError(error: Error): void

  type Wrapper<T extends AWSLambda.Handler> = (f: T) => T

  export function lambdaWrapper<T extends AWSLambda.Handler>(f: T): T
  export function stepLambdaWrapper<T extends AWSLambda.Handler>(f: T): T
  export function nodeWrapper<T extends Function>(f: T): T
  export function wrapBatchJob<T extends Function>(f: T): T
}

```

ranrib:
fix https://github.com/epsagon/epsagon-node/issues/100

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,23 @@
+// epsagon.d.ts
+declare module 'epsagon' {
+  import 'aws-lambda'
+
+  export function init(options: {
+    token: string
+    appName: string
+    metadataOnly?: boolean
+    useSSL?: boolean
+    traceCollectorURL?: string
+    isEpsagonDisabled?: boolean
+    urlPatternsToIgnore?: string[]
+  }): void
+  export function label(key: string, value: string): void
+  export function setError(error: Error): void
+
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -0,0 +1,23 @@
+// epsagon.d.ts
+declare module 'epsagon' {
+  import 'aws-lambda'
+
+  export function init(options: {
+    token: string
+    appName: string
```
ronyis, line 7: `appName` is basically optional (we use a default value)

ronyis (approved):

PR #110, opened 2019-06-20
feat(openwhisk): track API host and namespace for OpenWhisk invocations

trieloff:
Tracking the API host  can be useful for applications that are using multiple OpenWhisk instances, such as Adobe I/O Runtime and IBM BlueMix Functions together. Tracking the namespace is even more useful, because running one application across multiple namespaces (think user accounts) in the same instance is very common.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -988,7 +988,8 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
       "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=",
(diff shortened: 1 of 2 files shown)


trieloff:
@LiadGoren @ranrib I don't know if you have already established common names for the concepts:

`api_host` is the hostname that was used to invoke the action, for me it's typically either `runtime.adobe.io` or `adobeioruntime.net`, for IBM it will be something different. This might give you further differentiation in the UI for the type of function.

`namespace` is the user account, which is typically `trieloff` for my own testing and `helix` for stuff we have in production.

ranrib (approved):
💯

PR #113, opened 2019-06-24
provide an easy to of using OpenWhisk default parameters to initialize Epsagon

trieloff:
OpenWhisk does not provide an easy built-in way of setting environment vars, instead it it using a mechanism called [protected parameters](https://github.com/apache/incubator-openwhisk/blob/master/docs/webactions.md#protected-parameters) to enable operators to set parameters that will be used in every call to the action and that cannot be overwritten by users invoking the action. This makes it an accepted pattern to pass credentials and tokens as default parameters into the action.

Because `epsagon.init` was designed to be called before the wrapped function was invoked, implementing this pattern to use OpenWhisk default parameters with Epsagon required the user to write some boiler plate code like this:

```javascript
const epsagon = require('epsagon');


async function main(params) {
  // the main function
}

async function wrap(params) {
  epsagon.init({
    token: params.EPSAGON_TOKEN,
    appName: 'My App',
    metadataOnly: false, // Optional, send more trace data
  });

  return epsagon.openWhiskWrapper(main)(params);
}

module.exports.main = wrap;
```

With this PR, OpenWhisk users have a more convenient way of achieving the same goal, requiring less boilerplate code: simply pass the same `options` that `epsagon.init` takes as the second argument to `epsagon.openWhiskWrapper` and Epsagon will be initialized upon invocation of the action.

The more advanced (and better documented way) is to replace `token` with `token_param` in the `options` passed to `epsagon.openWhiskWrapper`, so that the token value can be taken from the specified parameter.


ranrib (approved):
I did some fixes around the CICD

PR #126, opened 2019-06-28
fix(config.js): user can set custom HTTP status as error

### LINKED ISSUE #124: Should OpenWhisk web actions that return an error HTTP status code be treated as errors?
I'm unsure if this is a bug, or what the desired behavior should be:

I have a function that will return a HTTP status code 404 if it can't find a particular file, e.g. by calling `https://adobeioruntime.net/api/v1/web/helix/helix-services/static@ci109?owner=trieloff&repo=helix-demo&ref=master&entry=/index.js&plain=true` and it returns 200 when everything is ok: `http "https://adobeioruntime.net/api/v1/web/helix/helix-services/static@ci109?owner=trieloff&repo=helix-demo&ref=master&entry=/index.js&plain=true&root=/htdocs"`

Both kinds of behavior are expected by the action developer, but not necessarily by the user of the action. 

For HTTP requests made by the function, Epsagon treats everything >=400 as an error. Should the same rule apply to HTTP responses returned by the action? 

I could also see that only HTTP status codes >=500 should be treated as an error in the action, and 4xx status code indicate an error only from the perspective of the caller, i.e. if it's called externally, then it's not an error, but if observed through #122 then it should be reported as an error, similar to a failed HTTP request.

ranrib:
can be set through init params, or environment variable:
```
epsagon.init({
    token: 'my-secret-token',
    appName: 'my-app-name',
    metadataOnly: true,
    httpErrorStatusCode: 500,
});
````
`EPSAGON_HTTP_ERR_CODE=500`

fixes https://github.com/epsagon/epsagon-node/issues/124

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -70,4 +70,9 @@ module.exports.setConfig = function setConfig(configData) {
     if (configData.urlPatternsToIgnore) {
         config.urlPatternsToIgnore = configData.urlPatternsToIgnore;
     }
+
+    // User-defined HTTP minimum status code to be treated as an error.
+    if (configData.httpErrorStatusCode) {
+        consts.HTTP_ERR_CODE = configData.httpErrorStatusCode;
+    }
 };
--- a/src/config.js
+++ b/src/config.js
@@ -3,6 +3,9 @@
  */
 const consts = require('./consts.js');
 
+// User-defined HTTP minimum status code to be treated as an error.
+module.exports.HTTP_ERR_CODE = parseInt(process.env.EPSAGON_HTTP_ERR_CODE, 10) || 400;
+
 /**
  * configuration singleton. preconfigured with default values.
  */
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/config.js [hunk 1]
```diff
@@ -70,4 +70,9 @@ module.exports.setConfig = function setConfig(configData) {
     if (configData.urlPatternsToIgnore) {
         config.urlPatternsToIgnore = configData.urlPatternsToIgnore;
     }
+
+    // User-defined HTTP minimum status code to be treated as an error.
+    if (configData.httpErrorStatusCode) {
+        consts.HTTP_ERR_CODE = configData.httpErrorStatusCode;
```
galbash: then it should be in config and not consts.

galbash (changes requested):

src/config.js
ranrib: You are right

galbash (dismissed):

galbash (approved):

ranrib:
@trieloff please update the lib at your time :)

PR #131, opened 2019-07-07
feat(utils): ignore fields in metadata

tvaintrob:


src/events/sql.js
```diff
@@ -1,5 +1,5 @@
 const uuid4 = require('uuid4');
-const { parse } = require('../resource_utils/sql_utils.js');
+const { parse } = require('../resource_utils/sqs_utils');
```
ranrib: Why?

tvaintrob:
had errors on that line trying to run the tests, said it couldn't find the
module that change fixed it

src/events/sql.js
tvaintrob: had errors on that line trying to run the tests, said it couldn't find the
module that change fixed it

ranrib (commented):
Looks good, please add an example to README

ranrib (approved):
💯

PR #133, opened 2019-07-10
fix(openwhisk): ensure the openwhisk action API is imported via require

### LINKED ISSUE #132: Openhwisk patcher doesn't work when final action is bundled
The problem is the:

```
```

```
let actions;
try {
} catch (e) {
    // ignore
}
```

tripodsan:
fixes #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,22 @@ function handler(event, context, callback) {
 }
 ```
 
+## Filter sensitive data
+
+You can pass a list of sensitive properties and they will be filtered out:
+
+```node
+epsagon.init({
+    token: 'my-secret-token',
+    appName: 'my-app-name',
+    metadataOnly: false, // Optional, send more trace data
+    ignoredKeys: ['password', ...]
+});
+```
+
+Alternatively you can pass a comma-separated list of sensitive keys using 
+the `EPSAGON_IGNORED_KEYS` environment variable to get the same effect.
+
 ## Web frameworks
 
 Support for Express, Hapi, and other frameworks is done through [epsagon-frameworks](https://github.com/epsagon/epsagon-node-frameworks)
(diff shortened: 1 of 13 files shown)


ranrib:
Hi @tripodsan, We use the tryRequire here (and on other libraries that we instrument) specifically to not force the load. We rely on that `openwhisk` is already provided by you at the packge.json. Isn't it that case?

ranrib:
Also, can you please fork out from `develop`? travis runs some acceptance test that only we can run, so we created develop for external contributions

tripodsan:
> Hi @tripodsan, We use the tryRequire here (and on other libraries that we instrument) specifically to not force the load. We rely on that `openwhisk` is already provided by you at the packge.json. Isn't it that case?

Hi @ranrib, the problem is that we use webpack to bundle our final openwhisk actions. using `tryRequire` hides the import of the `openwhisk/lib/action.js` from webpack and it is not possible to resolve its path. using a _normal_ require (which in turn uses the webpack require) solves this problem.

comparing the two bundles shows the following:

#### the espagon bundle
```diff
$ diff bundle.tryrequire.js bundle.require.js
15a16
> var actions = _interopDefault(require('openwhisk/lib/actions.js'));
13068c13069,13077
< const actions = try_require('openwhisk/lib/actions.js');
```

#### our bundle:
```diff
$ diff testing-bundle.require.js testing-bundle.tryrequire.js
12695a12696
> var actions = _interopDefault(__webpack_require__(/*! openwhisk/lib/actions.js */ "./node_modules/openwhisk/lib/actions.js"));
25748c25749,25757
< const actions = try_require('openwhisk/lib/actions.js');
```

The problem is that the `tryRequire()` gets transformed by webpack to:

```
    const currentRequire = req || __webpack_require__("./node_modules/epsagon/dist sync recursive");
```

so it only uses the module path of the `epsagon/dist` module.

tripodsan:
> Also, can you please fork out from develop

I would, but `develop` doesn't contain the latest code.

ranrib:
👍 taken care in #134

PR #135, opened 2019-07-11
Collect ecs metadata ep 1404

tvaintrob:


src/tracer.js
```diff
@@ -112,6 +113,11 @@ module.exports.addException = function addException(error, additionalData) {
 module.exports.initTrace = function initTrace(
     configData
 ) {
+    const ecsMetaUri = ecs.hasECSMetadata();
```
galbash: maybe do this on package import? people currently call init if they want to change the configuration

galbash (approved):

PR #137, opened 2019-07-14
fix(openwhisk): add support for concurrent tracers

ranrib:


src/tracer.js
```diff
@@ -203,11 +209,13 @@ function stripOperations(traceJson, attempt) {
  * handled
  * @param {function} traceSender: The function to use to send the trace. Gets the trace object
  *     as a parameter and sends a JSON version of it to epsagon's infrastructure
- * @param {Object} tracer  Optional tracer
  * @return {*} traceSender's result
  */
 function sendCurrentTrace(traceSender) {
     const tracerObj = module.exports.getTrace();
+    if (!tracerObj) {
+        return {};
```
galbash: I think this returns a promise, returning `Promise.resolve()` might be better

galbash (changes requested):
looks good, 1 comment inside

galbash (approved):

PR #138, opened 2019-07-14
feat(aws-sdk): adding custom promise support

galbash:
aws-sdk's code that overrides the promise function: 
image


razlo (approved):
maybe add a test for that?

PR #144, opened 2019-07-22
fix(openwhisk): ensure the openwhisk action API is imported via require

### LINKED ISSUE #132: Openhwisk patcher doesn't work when final action is bundled

tripodsan:
fixes #132


ranrib:
I'm trying something with the PR - closing for the test

PR #146, opened 2019-07-22
fix(openwhisk.js): support webpack

### LINKED ISSUE #132: Openhwisk patcher doesn't work when final action is bundled

ranrib:
fixes #132

### DIFF
--- a/src/events/openwhisk.js
+++ b/src/events/openwhisk.js
@@ -1,13 +1,20 @@
 const uuid4 = require('uuid4');
 const shimmer = require('shimmer');
-const tryRequire = require('../try_require.js');
 const utils = require('../utils.js');
 const tracer = require('../tracer.js');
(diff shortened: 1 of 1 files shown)


tripodsan:
I see that the acceptance tests for lambda are failing. I guess, the problem is that some 'packager' used for lambda is  trying to resolve all the `require`d modules. so the  `try/catch` in the openwhisk patcher doesn't work.

is there something you can do in the way how the lambda function is deployed?

otherwise I see 2 alternatives to this problem:

a) pass in the _api_ (i.e. the `actions.prototype`) in with the `openwhiskWrap()`, and init the wrapper only there.
b) (re)move the openwhisk support in an independent package (eg. `epsagon-openwhisk`) and add support for some plugin capability where the patchers are initialized.

@ranrib WDYT?

PR #149, opened 2019-07-28
fix(require): fixing tryRequire for webpack

galbash:
Forces webpack to use external require if not found in webpack require

### DIFF
--- a/src/try_require.js
+++ b/src/try_require.js
@@ -1,8 +1,12 @@
 let lastError = null;
 
-const tryRequire = (id, req) => {
+const tryRequire = (id) => {
     let path;
-    const currentRequire = req || require;
+    // for webpack we only support requiring external packages
+    const currentRequire = (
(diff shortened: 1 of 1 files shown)


tripodsan:
@galbash Nice find! Works for me.

galbash:
Great! I will update it today with some more content (currently it only finds moduled which are not bundled in webpack, so you won't be able to instrument http for exmaple, will try to find a fix for that)

galbash:
@tripodsan I updated the PR, currently (after a lot of webpack-wrestling) it should work, but we will only instrument external modules. So if you want to instrument pg, for example, you must specify it as external in your webpack config. Does that sound reasonable?

@ranrib would love for your take on this as well

ranrib (approved):

galbash:
@tripodsan just a head's up that the issue is merged to `master` and a version should be released soon :)

PR #151, opened 2019-07-31
fix(aws_sdk.py): fix sqs params and missing http headers

ranrib:
1. in SQS trigger we collect attributes and message attributes.
2. In SQS operation we collect message attributes.
3. In HTTP event we will re-collect request headers again.

### DIFF
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -135,6 +135,7 @@ function httpWrapper(wrappedFunction) {
             const url = `${protocol}://${hostname}${pathname}`;
             const fullURL = `${url}${path}`;
             httpEvent.setResource(resource);
+
             eventInterface.addToMetadata(httpEvent,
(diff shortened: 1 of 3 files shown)


src/events/http.js
```diff
@@ -158,6 +159,16 @@ function httpWrapper(wrappedFunction) {
                     response_headers: res.headers,
                 });
 
+                // Override request headers if they are present here. In some libs they are not
+                // available on `options.headers`
+                // eslint-disable-next-line no-underscore-dangle
+                if (res.req && res.req._headers) {
+                    eventInterface.addToMetadata(httpEvent, metadataFields, {
```
galbash, line 166: Is this tge first time you are adding metadata fields or is this redundant?

galbash (approved):

Issue #127, opened 2019-06-28
Provide Option to Mask Sensitive Data


trieloff:
When functions carry sensitive data such as API tokens, E-Mail addresses, etc. it is desirable to keep them away from logging infrastructure.

Therefore, I propose the addition of two new parameters for `options`: `mask_fields` and `mask_values` that specifies each a list of strings or regular expressions that will be applied to traces in `sendCurrentTrace`. 

```javascript
epsagon.init({
  // the usual stuff omitted
  mask_fields: ['EPSAGON_TOKEN', 'Authorization'], // mask well-known field and header names
  mask_values: [/[0-9a-zA-Z/+]{40}/, /[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}/] // mask AWS Secret Keys and E-Mail addresses 
});
```

Initially, I'd just mask things with `--------------`, but I could imagine a more advanced hash-based masking that would enable search in the Epsagon UI (by applying the same hash to the search term).

ranrib:
It's not the exact thing - but it might help you for the meantime https://github.com/epsagon/epsagon-node/pull/131

trieloff:
I'm reopening this. When you add package-level parameters, it becomes very hard to manage this without wildcards.

ranrib:
I think it's solved and published already, so I'm closing

PR #164, opened 2019-08-26
feat(config): ignoredKeys supports regular expressions

trieloff:
Simple string-based comparison is often not enough, and this change allows pattern-based surpression of keys

partial fix of #127


ranrib (approved):
Awesome work 💯

PR #168, opened 2019-08-27
fix(http.js): multiple params not supported

### LINKED ISSUE #165: http.get(url, options) not supported
The `get` function of the NodeJS `http` module can be called with `options` or with `url, [options]. The wrapper was built only for the `options` case, and assumes that the first argument is always an object, when it can be an object, a string, or a `URL` object.

When `http.get` is used with a URL as in https://github.com/adobe/helix-resolve-git-ref/blob/233400fb22d69c79153d7ff126d8ca85c0210ce2/src/index.js#L44 it results in traces to HTTP requests being recorded as pointing to `http://localhost/`. This is due to the assumption that `options` will always be an object here: https://github.com/epsagon/epsagon-node/blob/b9ab8dd8043087a34abe2bcd0c69a669ea6fe51b/src/events/http.js#L90-L95

henperez:
Fixed #165

### DIFF
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -81,7 +81,7 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
  * @param {callback} callback The callback function, if exists
  * @returns {object} The params array
  */
-    function buildParams(input, options, callback) {
+function buildParams(input, options, callback) {
     if (input && options) {
         // In case of both input and options returning all three
         return [input, options, callback];
@@ -102,8 +102,8 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
 function httpWrapper(wrappedFunction) {
     return function internalHttpWrapper(a, b, c) {
         let input = a;
-        let options = b
-        let callback = c
+        let options = b;
+        let callback = c;
         if (!(['string', 'URL'].includes(typeof input)) && !callback) {
             callback = b;
             options = a;
@@ -162,8 +162,8 @@ function httpWrapper(wrappedFunction) {
                 return wrappedFunction.apply(this, [a, b, c]);
             }
 
-            // eslint-disable-next-line no-underscore-dangle
             const agent = (
+                // eslint-disable-next-line no-underscore-dangle
                 (options && options.agent) || (options && options._defaultAgent) ||
                 undefined
             );
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -6,6 +6,7 @@ const uuid4 = require('uuid4');
 const shimmer = require('shimmer');
 const http = require('http');
 const https = require('https');
+const urlLib = require('url');
 const tryRequire = require('../try_require.js');
 const utils = require('../utils.js');
 const tracer = require('../tracer.js');
@@ -13,7 +14,6 @@ const serverlessEvent = require('../proto/event_pb.js');
 const eventInterface = require('../event.js');
 const errorCode = require('../proto/error_code_pb.js');
 const config = require('../config.js');
-const url = require('url');
 
 const Wreck = tryRequire('wreck');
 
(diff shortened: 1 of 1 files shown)


src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
 }
 
+/**
+ * Builds the HTTP Params array
+ * @param {string} input The URL, if exists
+ * @param {object} options The Options object, if exists
+ * @param {callback} callback The callback function, if exists
+ * @returns {object} The params array
+ */
+function buildParams(input, options, callback) {
```
galbash: maybe use ...args or arg1, arg2, arg3

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+            options = a;
+            input = undefined;
+        }
+
+        if ((typeof options === 'function') && (!callback)) {
+            callback = options;
+            options = null;
+        }
+
         if (callback && callback.__epsagonCallback) { // eslint-disable-line no-underscore-dangle
```
galbash, line 120: move this to before the args parsing

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+
 /**
  * Wraps the http's module request function with tracing
  * @param {Function} wrappedFunction The http's request module
  * @returns {Function} The wrapped function
  */
 function httpWrapper(wrappedFunction) {
-    return function internalHttpWrapper(options, callback) {
+    return function internalHttpWrapper(a, b, c) {
+        let input = a;
```
galbash: call it url (like the API?)

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+}
+
 /**
  * Wraps the http's module request function with tracing
  * @param {Function} wrappedFunction The http's request module
  * @returns {Function} The wrapped function
  */
 function httpWrapper(wrappedFunction) {
-    return function internalHttpWrapper(options, callback) {
+    return function internalHttpWrapper(a, b, c) {
```
galbash: instead of a, b, c you can use `...args` I think

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+        let input = a;
+        let options = b;
+        let callback = c;
+        if (!(['string', 'URL'].includes(typeof input)) && !callback) {
```
galbash: document (here is `request(options, [callback])`)

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+            callback = b;
+            options = a;
+            input = undefined;
+        }
+
+        if ((typeof options === 'function') && (!callback)) {
```
galbash, line 115: doc this case as well

src/events/http.js
```diff
@@ -73,51 +74,114 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+
+            if (typeof parsedUrl === 'string') {
+                parsedUrl = urlLib.parse(parsedUrl);
+            }
+
             const hostname = (
-                options.hostname ||
-                options.host ||
-                (options.uri && options.uri.hostname) ||
+                (parsedUrl && parsedUrl.hostname) ||
```
galbash, line 134: did you check this? should work It just seems like maybe it won't

galbash (changes requested):
looks good, few comments

galbash (approved):

Issue #178, opened 2019-09-27
Provide option to see queries when sending requests to BigQuery


MarquiseRosier:
it'd be pretty cool if we can see the query being sent to bigquery in the tags!

trieloff:
Ideally it would work similar to the PG wrapper: https://github.com/epsagon/epsagon-node/blob/master/src/events/pg.js

ranrib:
Added in https://github.com/epsagon/epsagon-node/issues/178

PR #181, opened 2019-10-22
feat(google_cloud): instrumentation for google bigquery

liadgo:


src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+const tryRequire = require('../try_require.js');
+const utils = require('../utils.js');
+const tracer = require('../tracer.js');
+const serverlessEvent = require('../proto/event_pb.js');
+const eventInterface = require('../event.js');
+const errorCode = require('../proto/error_code_pb.js');
+
+const common = tryRequire('@google-cloud/common/');
+
+const uriSplitString = 'googleapis.com/';
```
galbash: Const should be all caps i think (look at other modules

src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+ * @returns {Function} The wrapped function
+ */
+function googleWrapper(wrappedFunction) {
+    return function internalOWWrapper(reqOpts, config, callback) {
+        const uri = reqOpts.uri.split(uriSplitString)[1];
+        const splitUri = uri.split('/');
+        const service = splitUri[0];
+        const projectId = splitUri[3];
+        const path = uri.split(`${projectId}/`)[1];
+        const operation = path.split('/')[0];
```
galbash: Al these assignments seems very opportunistic (past  7+- lines), nothing can go wrong here? Do we want some default values?

src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+        const resource = new serverlessEvent.Resource([
+            projectId,
```
galbash, line 31: Is this the resource name? We probably want a better name

src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+            service,
+            operation,
+        ]);
+        const startTime = Date.now();
+        const invokeEvent = new serverlessEvent.Event([
+            `${service}-${uuid4()}`,
```
galbash: Is there no google messagi id or something? Per service, might be useful for later strong id connections

src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+            service,
+            0,
+            errorCode.ErrorCode.OK,
+        ]);
+
+        invokeEvent.setResource(resource);
+        if (reqOpts.json !== undefined) {
+            eventInterface.addToMetadata(
+                invokeEvent,
+                reqOpts.json
```
galbash: Do it as the next parameter maybe? Not to add on metadata only

src/events/google_cloud.js
```diff
@@ -0,0 +1,77 @@
...
+            );
+        }
+
+        let patchedCallback;
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, body, response) => {
+                invokeEvent.setDuration(utils.createDurationTimestamp(startTime));
+                eventInterface.addToMetadata(invokeEvent, response.body);
+                callback(err, body, response);
+                resolve();
```
galbash, line 68: Resolve before calling the callback (or we might never resolve on exceptions)

galbash (changes requested):
Looks really good, few small comments

src/events/google_cloud.js
liadgo: K

src/events/google_cloud.js
liadgo: OK

src/events/google_cloud.js
liadgo: :)

src/events/google_cloud.js
liadgo: Talked about it

galbash (dismissed):

razlo (approved):

ranrib:
fixes https://github.com/epsagon/epsagon-node/issues/178

PR #189, opened 2019-11-04
fix(http): remove response body collection

liadgo:


src/events/http.js
```diff
@@ -60,7 +60,7 @@ function isBlacklistHeader(headers) {
  * @param {string} url The URL to check
  * @returns {boolean} True if it is in the user-defined blacklist, False otherwise.
  */
-function isURLIgnoredByUser(url) {
+function isURLIgnoredByUser(url) { // eslint-disable-line no-unused-vars
```
galbash, line 63: why is the linting disable required here? you use url

galbash: you can remove it if it is not used

galbash (approved):

PR #194, opened 2019-11-18
feat(tracer.js): toggle epsagon

henperez:


src/tracer.js
```diff
@@ -405,7 +405,7 @@ module.exports.postTrace = function postTrace(traceObject) {
  */
 module.exports.sendTrace = function sendTrace(runnerUpdateFunc) {
     const tracerObj = module.exports.getTrace();
-    if (!tracerObj) {
+    if (!tracerObj || tracerObj.disabled === true) {
```
ranrib: `tracerObj.disabled === true` -> `tracerObj.disabled`

src/tracer.js
```diff
@@ -425,7 +425,7 @@ module.exports.sendTrace = function sendTrace(runnerUpdateFunc) {
  */
 module.exports.sendTraceSync = function sendTraceSync() {
     const tracerObj = module.exports.getTrace();
-    if (!tracerObj || tracerObj.disabled === true) {
+    if (!tracerObj || tracerObj.disabled) {
```
ranrib: This is not a good place to test, because if tracerObj is undefined you'll get an exception

ranrib (approved):

PR #198, opened 2019-11-24
fix(tracer.js): trim big traces

henperez:


src/wrappers/node.js
```diff
@@ -70,6 +73,22 @@ module.exports.nodeWrapper = function nodeWrapper(functionToWrap) {
...
+            if (!getConfig().metadataOnly) {
+                let jsonResult;
+                try {
+                    jsonResult = JSON.stringify(
+                        typeof result === 'undefined' ? null : result
+                    );
+                } catch (err) {
+                    jsonResult = `${FAILED_TO_SERIALIZE_MESSAGE}: ${err.message}`;
+                }
+                eventInterface.addToMetadata(
```
adavidai, line 85: Is return value part of metadata?

adavidai (approved):

PR #203, opened 2019-12-19
feat(dns): add support for dns

idonava:
[https://epsagon.atlassian.net/browse/EP-1838](https://epsagon.atlassian.net/browse/EP-1838)
`process.env.DNS_INSTRUMENTATION=true` for start tracing dns


examples/dns_example.js
```diff
@@ -0,0 +1,122 @@
+const epsagon = require('../src/index.js');
```
ranrib: Ignore this file

idonava: file removed.

ranrib (dismissed):

ranrib:
💯

.gitignore
```diff
@@ -1,4 +1,5 @@
 .idea/
 node_modules/
 dist/
-src/resource_utils/sql_utils.js
\ No newline at end of file
+src/resource_utils/sql_utils.js
+**/.DS_Store
```
galbash: add newline

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+const URL_BLACKLIST = {
+    'tc.epsagon.com': 'endsWith',
+};
+/**
+ * Checks if a URL is in the blacklist
+ * @param {string} url The URL to check
+ * @param {string} path The Path to check (optional)
+ * @returns {boolean} True if it is in the blacklist, False otherwise
+ */
+function isBlacklistURL(url, path) {
```
galbash: maybe move to a common helper? (not sure it's the same function signature but if it is i'd recommend it)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    return Object.keys(URL_BLACKLIST).some((key) => {
+        if (typeof URL_BLACKLIST[key] === typeof (() => {})) {
+            return URL_BLACKLIST[key](url, key, path);
+        }
+        return url[URL_BLACKLIST[key]](key);
+    });
+}
+
+const rrtypesMethods = {
```
galbash, line 16: rr? add doc please

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    resolveMx: 'MX',
+    resolveNaptr: 'NAPTR',
+    resolveNs: 'NS',
+    resolvePtr: 'PTR',
+    resolveSoa: 'SOA',
+    resolveSrv: 'SRV',
+    resolveTxt: 'TXT',
+};
+
+const initialDnsEvent = (name) => {
```
galbash, line 35: add doc to the functions (fix everywhere)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+        null,
+        'dns',
+        0,
+        errorCode.ErrorCode.OK,
+    ]);
+    dnsEvent.setResource(resource);
+    return { dnsEvent, startTime };
+};
+
+const endDurationAndAddCallbackMetadata = (dnsEvent, startTime, err, metadata) => {
```
galbash: finalizeEvent? you can add in the doc the specific actions

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    dnsEvent.setDuration(utils.createDurationTimestamp(startTime));
+    if (err) {
+        eventInterface.setException(dnsEvent, err);
+    } else {
+        eventInterface.addToMetadata(dnsEvent, { ...metadata });
```
galbash: are you sure you want to add the entire metadata? maybe do it only when metadataOnly=False (third argument instead of second)? consult with @henprtz

galbash: also no need to re-spread metadata i guess?

galbash: also make sure the naming and size of the response is reasonable. whitelisting what we add to the metadata is usually the better approach

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+
+const getRrtypeArguments = (arg1, arg2, arg3, functionName) => {
+    const hostname = arg1;
+    let rrtype = arg2;
+    let callback = arg3;
+    if (!arg3) {
+        if (functionName) {
+            rrtype = Object.values(rrtypesMethods).find(type => functionName.toLocaleLowerCase().includes((`query${type}`.toLocaleLowerCase())));
+        }
+        if (!rrtype) rrtype = rrtypesMethods.resolve4;
```
galbash: always use {}. fix everywhere

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+const endDurationAndAddCallbackMetadata = (dnsEvent, startTime, err, metadata) => {
+    dnsEvent.setDuration(utils.createDurationTimestamp(startTime));
+    if (err) {
+        eventInterface.setException(dnsEvent, err);
+    } else {
+        eventInterface.addToMetadata(dnsEvent, { ...metadata });
+    }
+};
+
+const getRrtypeArguments = (arg1, arg2, arg3, functionName) => {
```
galbash, line 82: add the unroll logic to doc here

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    let options = arg2;
+    let callback = arg3;
+    if (!arg3) {
+        options = undefined;
+        callback = arg2;
+    }
+    return { hostname, options, callback };
+};
+
+const buildParams = (arg1, arg2, arg3) => {
```
galbash: you can just always pass `[arg1, arg2, arg3]`, this function is redundent

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    if (!arg3) {
+        return [arg1, arg2];
+    }
+    return [arg1, arg2, arg3];
+};
+
+const wrapDnsResolveFunction = original => (arg1, arg2, arg3) => {
```
galbash: I recall there were some issues when using arrow functions with shimmer (that's why most of this project doesn't use them) - something that has to do with losing this. let's talk about it on sunday.

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    let patchedCallback;
+    let clientRequest;
+    let options;
+    const { hostname, rrtype, callback } = getRrtypeArguments(arg1, arg2, arg3, original.name);
+    if (!callback) return original.apply(this, buildParams(arg1, arg2, arg3));
```
galbash: use {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    if (typeof arg2 === 'object') options = arg2;
```
galbash: you have like 3 unrolling functions, use one of them. also {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    try {
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        let requestArg2;
+        if (options) {
+            requestArg2 = { options };
```
galbash: the keys in the options object probably won't make good metadata names. let's go over it on Sunday and see if we really need it all (remember the user request was to find timing effects better)

src/events/dns.js
galbash: wait also why? we can still have our callback set up and wait for the response no?

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+        } else {
+            requestArg2 = { 'Resource record type': rrtype };
+        }
+        eventInterface.addToMetadata(dnsEvent, { hostname, ...requestArg2 });
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, arg) => {
```
galbash: arg -> result (unless the documentation explicitly calls it arg)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+                let callbackArg;
+                if (options) {
```
galbash: add a bit of doc to why if options exists you use 'addresses'

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+                    callbackArg = { addresses: arg };
+                } else {
+                    callbackArg = { records: arg };
+                }
+                endDurationAndAddCallbackMetadata(dnsEvent, startTime, err, { ...callbackArg });
+                resolve();
+                if (callback) callback(err, arg);
```
galbash: {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+            };
+        });
+        clientRequest = original.apply(this, [hostname, rrtype, patchedCallback]);
```
galbash: just making sure that this always return a value even if callback was passed (otherwise you will invoke the original method twice). Please show it to me on sunday.

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    if (!clientRequest) {
+        clientRequest = original.apply(this, buildParams(arg1, arg2, arg3));
+    }
+    return clientRequest;
+};
+
+const wrapDnsLookupServiceFunction = original => (address, port, callback) => {
+    let patchedCallback;
+    let clientRequest;
+    if (!callback) return original.apply(this, buildParams(address, port, callback));
```
galbash: {}

galbash: also why? still add a callback even if the user did not supply one

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    try {
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        eventInterface.addToMetadata(dnsEvent, { address, port });
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, hostname, service) => {
+                endDurationAndAddCallbackMetadata(dnsEvent, startTime, err, { hostname, service });
+                resolve();
+                if (callback) callback(err, hostname, service);
```
galbash: {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+            };
+        });
+        clientRequest = original.apply(this, [address, port, patchedCallback]);
```
galbash: same (making sure it returns a value)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    let patchedCallback;
+    let clientRequest;
+    try {
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        eventInterface.addToMetadata(dnsEvent, { ip });
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, hostnames) => {
+                endDurationAndAddCallbackMetadata(dnsEvent, startTime, err, { hostnames });
+                resolve();
+                if (callback) callback(err, hostnames);
```
galbash: {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+            };
+        });
+        clientRequest = original.apply(this, [ip, patchedCallback]);
```
galbash: same (making sure it returns a value)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    if (!clientRequest) {
+        clientRequest = original.apply(this, [ip, callback]);
+    }
+    return clientRequest;
+};
+
+const wrapDnsLookupFunction = original => (arg1, arg2, arg3) => {
+    let patchedCallback;
+    let clientRequest;
+    const { hostname, options, callback } = getArguments(arg1, arg2, arg3);
```
galbash: if getArguments is specific to lookup, it should have a better name.

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    if (isBlacklistURL(hostname)) {
+        utils.debugLog(`filtered blacklist hostname ${hostname}`);
+        return original.apply(this, buildParams(arg1, arg2, arg3));
+    }
+    try {
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        eventInterface.addToMetadata(dnsEvent, { hostname });
+        if (options) eventInterface.addToMetadata(dnsEvent, { options });
```
galbash: {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, address, family) => {
+                endDurationAndAddCallbackMetadata(dnsEvent, startTime, err, { address, family });
+                resolve();
+                if (callback) callback(err, address, family);
```
galbash: {}

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    return clientRequest;
+};
+
+const wrapDnsLookupFunction = original => (arg1, arg2, arg3) => {
+    let patchedCallback;
+    let clientRequest;
+    const { hostname, options, callback } = getArguments(arg1, arg2, arg3);
+    if (isBlacklistURL(hostname)) {
+        utils.debugLog(`filtered blacklist hostname ${hostname}`);
+        return original.apply(this, buildParams(arg1, arg2, arg3));
```
galbash: just FYI, you can use the `arguments` keyword here and in similar places (but the rest of the instrumentation doesn't so take the suggestion I put on the actual buildParams method instead, just wanted you to be acquainted with it)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+        eventInterface.addToMetadata(dnsEvent, { hostname });
+        if (options) eventInterface.addToMetadata(dnsEvent, { options });
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, address, family) => {
+                endDurationAndAddCallbackMetadata(dnsEvent, startTime, err, { address, family });
+                resolve();
+                if (callback) callback(err, address, family);
+            };
+        });
+        clientRequest = original.apply(this, buildParams(hostname, options, patchedCallback));
```
galbash: same (make sure returns value always)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+    return clientRequest;
+};
+
+module.exports = {
+    /**
+     * Initializes the dns tracer
+     */
+    init() {
+        // eslint-disable-next-line no-debugger
+        debugger;
```
galbash: please remove...

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+        if ((process.env.DNS_INSTRUMENTATION || '').toUpperCase() === 'TRUE') {
```
galbash: I think `if (process.env.EPSAGON_INSTRUMENTATION_DNS)` is enough here (no need to make sure the content, also add EPSAGON prefix)

src/events/dns.js
```diff
@@ -0,0 +1,239 @@
...
+            Object.keys(rrtypesMethods).forEach((functionToTrace) => {
```
galbash, line 344: did you look at the implementation? are you 100% sure they are not calling the same function underneath and we have to patch them all? (an argument of "they might but patching them all is easy and is less prone to break because of implementation changes" might be valid here, just making sure this is your thought process)

galbash (changes requested):
Overall looks good. Few changes required especially around:
1. documentation, code conventions & style
2. no-harm verifications

galbash:
also we need to add the flag to the docs somewhere (readme / docs on site)

src/events/dns.js
idonava: Nice, I added helpers/events.js and converted http&dns events to use it.

src/events/dns.js
idonava: > are you sure you want to add the entire metadata? maybe do it only when metadataOnly=False (third argument instead of second)? consult with @henprtz

At a minimum (and most commonly) is a string/integer and at maximum it could be an array/object of strings (https://nodejs.org/api/dns.html). I think it's relevant data for those who decide to trace dns.
example for the longest metadata I got so far:
```
{
"retry":3600
"expire":604800
"minttl":3601
"nsname": "dns1.registrar-servers.com" 
"serial":1573449957
"refresh":43200
"hostmaster": "hostmaster.registrar-servers.com" 
}
```
Hostname: e


> also no need to re-spread metadata i guess?

It's not necessary after a small change i made in the code, tnx.



> also make sure the naming and size of the response is reasonable. whitelisting what we add to the metadata is usually the better approach

@galbash There is any size limitation of the metadata i need to know?

src/events/dns.js
idonava: I didn't understand where i need use {}

src/events/dns.js
idonava: you right for the most calls instead of one, I deleted the function and fix the one specific in the code.

src/events/dns.js
idonava: I think it can cause a bug.
When I read process.env.DNS_INSTRUMENTATION I getting the value as a string.
And every value I set to  DNS_INSTRUMENTATION will pass that way.

src/events/dns.js
idonava: All those functions (rrtypesMethods keys) calls to their own function and i catch them all in dns.Resolve, I'm checking now if dns library contains each function and just then wraps it with shimmer.

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+    resolveNaptr: 'NAPTR',
+    resolveNs: 'NS',
+    resolvePtr: 'PTR',
+    resolveSoa: 'SOA',
+    resolveSrv: 'SRV',
+    resolveTxt: 'TXT',
+};
+
+/**
+ * Initial Dns event.
```
galbash: creates and initial DNS event in the protobuf format (you just wrote the function name..)

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+    }
+    dnsEvent.setDuration(utils.createDurationTimestamp(startTime));
+};
+
+/**
+ * Checking resource record type by resolve function name (default resolve4 - 'A').
+ * @param {*} arg1 function first argument.
+ * @param {*} arg2 function second argument
+ * @param {*} arg3 function third argument
+ * @param {*} functionName function resolve name.
```
galbash: string?

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+ * @returns {Object} Object of Hostname to resolve, Resource record type, and callback.
+ */
+const getRrtypeArguments = (arg1, arg2, arg3, functionName) => {
+    const hostname = arg1;
+    let rrtype = arg2;
+    let callback = arg3;
+    // in case of arg3 doesnt exist or arg2 is options.
+    if (!arg3 || typeof arg2 === 'object') {
+        if (functionName) {
```
galbash, line 88: I think you can use a default value for the parameter instead?

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+        }
+        if (!arg3) {
+            callback = arg2;
+        }
+    }
+    return { hostname, rrtype, callback };
+};
+
+/**
+ * Gettign lookup arguments with options.
```
galbash: typo

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+ * @param {*} arg1 function first argument.
+ * @param {*} arg2 function second argument
+ * @param {*} arg3 function third argument
+ * @returns {Object} Object of Hostname, options (if doesnt exist- undifiend) and callback.
```
galbash: typo in undefined

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+
+/**
+ * Gettinng callback object of dns resolve.
+ * @param {*} arg callback argument
+ * @param {string} rrtype resouece record type
+ * @returns {Object} Returns the callback object with the appropriate key compared to rrtype,
+ * (if doesnt exist - records).
+ */
+const getCallbackResolveArgument = (arg, rrtype) => {
+    if (!arg) return undefined;
```
galbash: always {}

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+    let callbackArg;
```
galbash, line 127: I think you can just return in the ifs instead of having this vars (not a must, whichever you prefer)

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+/**
+ * Calling to the dns function without callback, And record the error if thrown.
+ * @param {Function} original dns function.
+ * @param {number} startTime Event start time.
+ * @param {serverlessEvent.Event} dnsEvent Dns event.
+ * @param {Array} args Array of function arguments.
+ * @returns {Object} original function response.
+ */
+function handleFunctionWithoutCallback(original, startTime, dnsEvent, args) {
+    let clientRequest;
```
galbash: again, you can just call return in the `try` block (so this var is redundant)

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+        let patchedCallback;
+        let clientRequest;
+        let options;
+        const { hostname, rrtype, callback } = getRrtypeArguments(arg1, arg2, arg3, original.name);
+        if (typeof arg2 === 'object') {
+            options = arg2;
+        }
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        if (!callback) {
+            clientRequest = handleFunctionWithoutCallback(
```
galbash: I would just make this `return handleFunctionWithoutCallback` so it's obvious that execution doesn't continue after this point

galbash: this way you don't need the `else` block

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+ * @param {Function} original The dns function.
+ * @returns {Function} The wrapped function
+ */
+function wrapDnsLookupServiceFunction(original) {
+    return function internalWrapDnsLookupServiceFunction(address, port, callback) {
+        let patchedCallback;
+        let clientRequest;
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        if (!callback) {
+            clientRequest = handleFunctionWithoutCallback(
```
galbash: use return here as well

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+ * @param {Function} original The dns function.
+ * @returns {Function} The wrapped function
+ */
+function wrapDnsReverseFunction(original) {
+    return function internalWrapDnsReverseFunction(ip, callback) {
+        let patchedCallback;
+        let clientRequest;
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        if (!callback) {
+            clientRequest = handleFunctionWithoutCallback(
```
galbash: same

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+};
+
+/**
+ * Adding callback data/error to event, and finalize event.
+ * @param {serverlessEvent.Event} dnsEvent Dns event.
+ * @param {number} startTime Event start time.
+ * @param {Error} err Callback error.
+ * @param {string[] | Object[] | Object} metadata Callback metadata.
+ */
+const finalizeEvent = (dnsEvent, startTime, err, metadata) => {
```
galbash: maybe try catch this function and call `tracer.addException` if an error occurs (so we are more robust and won't disrupt user's callback)

src/events/dns.js
```diff
@@ -0,0 +1,358 @@
...
+        let patchedCallback;
+        let clientRequest;
+        const { hostname, options, callback } = getLookupArguments(arg1, arg2, arg3);
+        if (isBlacklistURL(hostname, URL_BLACKLIST)) {
+            utils.debugLog(`filtered blacklist hostname ${hostname}`);
+            return original.apply(this, [arg1, arg2, arg3]);
+        }
+        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        if (!callback) {
+            clientRequest = handleFunctionWithoutCallback(
```
galbash: same here (return)

src/helpers/events.js
```diff
@@ -0,0 +1,29 @@
+
+/**
+ * Checks if a URL is in the blacklist
+ * @param {string} url The URL to check
+ * @param {object} URL_BLACKLIST Object of blacklist url objects (KEY=[url], VALUE=[condition]).
+ * @param {string} path The Path to check (optional)
+ * @returns {boolean} True if it is in the blacklist, False otherwise
+ */
+const isBlacklistURL = (url, URL_BLACKLIST, path) => Object.keys(URL_BLACKLIST).some((key) => {
```
galbash: parameter names should be lowercased

src/helpers/events.js
```diff
@@ -0,0 +1,29 @@
...
+    return url[URL_BLACKLIST[key]](key);
+});
+
+/**
+ * Checks if a user agent header is in the blacklist
+ * @param {string} headers The Headers to check.
+ * @param {Array} USER_AGENTS_BLACKLIST Array of blacklist user agents.
+ * @returns {boolean} True if it is in the blacklist, False otherwise
+ */
+const isBlacklistHeader = (headers, USER_AGENTS_BLACKLIST) => {
```
galbash: same (also fix doc)

galbash (changes requested):
few small style comments

src/events/dns.js
idonava: nice

galbash (approved):

PR #208, opened 2020-01-07
feat(pubsub): support pubsub

idonava:
Added google pubsub instrumentation

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -2,12 +2,14 @@
  * @fileoverview an interface to {@link proto.event_pb.Event} objects, with useful methods to
  * manipulate them
  */
+const uuid4 = require('uuid4');
 const utils = require('./utils');
 const errorCode = require('./proto/error_code_pb.js');
 const exception = require('./proto/exception_pb.js');
 const config = require('./config.js');
 const tracer = require('./tracer.js');
 const consts = require('./consts.js');
+const serverlessEvent = require('./proto/event_pb.js');
 
 /**
  * Sets an event's exception to the given error
@@ -101,5 +103,53 @@ module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, valu
     const labelsJson = JSON.stringify(labels);
     if (labelsJson.length <= consts.MAX_LABEL_SIZE) {
         event.getResource().getMetadataMap().set('labels', labelsJson);
+    }
+};
+
+/**
+ * Create and initialize a new serverless event in the epsagon format.
+ * @param {string} resourceType resourceType name
+ * @param {string} name Event name
+ * @param {string} operation Operation name
+ * @param {string} origin Origin name (optional)
+ * @returns {Object} Object with serverlessEvent and event started time.
+ */
+module.exports.initializeEvent = (resourceType, name, operation, origin) => {
+    const startTime = Date.now();
+    const resource = new serverlessEvent.Resource([
+        name,
+        resourceType,
+        operation,
+    ]);
+    const slsEvent = new serverlessEvent.Event([
+        `${resourceType}-${uuid4()}`,
+        utils.createTimestampFromTime(startTime),
+        null,
+        origin || resourceType,
+        0,
+        errorCode.ErrorCode.OK,
+    ]);
+    slsEvent.setResource(resource);
+    return { slsEvent, startTime };
+};
+
+/**
+ * Adding callback data/error to event, and finalize event.
+ * @param {serverlessEvent.Event} slsEvent Serverless event.
+ * @param {number} startTime Event start time.
+ * @param {Error} error Callback error.
+ * @param {string[] | Object[] | Object} metadata Callback metadata.
+ */
+module.exports.finalizeEvent = (slsEvent, startTime, error, metadata) => {
+    try {
+        if (error) {
+            this.setException(slsEvent, error);
+        }
+        if (metadata) {
+            this.addToMetadata(slsEvent, metadata);
+        }
+        slsEvent.setDuration(utils.createDurationTimestamp(startTime));
+    } catch (err) {
+        tracer.addException(err);
     }
 };
--- a/src/event.js
+++ b/src/event.js
@@ -111,7 +111,7 @@ module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, valu
  * @param {string} resourceType resourceType name
  * @param {string} name Event name
  * @param {string} operation Operation name
- * @param {string} origin Origin name (optional)
+ * @param {string} origin Origin name
  * @returns {Object} Object with serverlessEvent and event started time.
  */
 module.exports.initializeEvent = function initializeEvent(resourceType, name, operation, origin) {
@@ -125,7 +125,7 @@ module.exports.initializeEvent = function initializeEvent(resourceType, name, op
         `${resourceType}-${uuid4()}`,
         utils.createTimestampFromTime(startTime),
         null,
-        origin || resourceType,
+        origin,
         0,
         errorCode.ErrorCode.OK,
     ]);
--- a/src/event.js
+++ b/src/event.js
@@ -114,7 +114,7 @@ module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, valu
  * @param {string} origin Origin name (optional)
  * @returns {Object} Object with serverlessEvent and event started time.
  */
-module.exports.initializeEvent = (resourceType, name, operation, origin) => {
+module.exports.initializeEvent = function initializeEvent(resourceType, name, operation, origin) {
     const startTime = Date.now();
     const resource = new serverlessEvent.Resource([
         name,
@@ -140,7 +140,7 @@ module.exports.initializeEvent = (resourceType, name, operation, origin) => {
  * @param {Error} error Callback error.
  * @param {string[] | Object[] | Object} metadata Callback metadata.
  */
-module.exports.finalizeEvent = (slsEvent, startTime, error, metadata) => {
+module.exports.finalizeEvent = function finalizeEvent(slsEvent, startTime, error, metadata) {
     try {
         if (error) {
             this.setException(slsEvent, error);
--- a/src/events/dns.js
+++ b/src/events/dns.js
@@ -3,7 +3,7 @@ const shimmer = require('shimmer');
 const utils = require('../utils.js');
 const tracer = require('../tracer.js');
 const eventInterface = require('../event.js');
-const { isBlacklistURL, initializeEvent, finalizeEvent } = require('.././helpers/events');
+const { isBlacklistURL } = require('.././helpers/events');
 
 const URL_BLACKLIST = {
     'tc.epsagon.com': 'endsWith',
@@ -123,7 +123,7 @@ function wrapDnsResolveFunction(original) {
         if (typeof arg2 === 'object') {
             options = arg2;
         }
-        const { event: dnsEvent, startTime } = initializeEvent('dns', original.name, 'dns');
+        const { slsEvent: dnsEvent, startTime } = eventInterface.initializeEvent('dns', original.name, 'dns');
         if (!callback) {
             return handleFunctionWithoutCallback(
                 original, startTime, dnsEvent, [arg1, arg2, arg3]
@@ -141,7 +141,7 @@ function wrapDnsResolveFunction(original) {
             const responsePromise = new Promise((resolve) => {
                 patchedCallback = (err, records) => {
                     const callbackArgument = getCallbackResolveArgument(records, rrtype);
-                    finalizeEvent(dnsEvent, startTime, err, { ...callbackArgument });
+                    eventInterface.finalizeEvent(dnsEvent, startTime, err, { ...callbackArgument });
                     resolve();
                     if (callback) {
                         callback(err, records);
@@ -168,17 +168,17 @@ function wrapDnsLookupServiceFunction(original) {
     return function internalWrapDnsLookupServiceFunction(address, port, callback) {
         let patchedCallback;
         let clientRequest;
-        const { event, startTime } = initializeEvent('dns', original.name, 'dns');
+        const { slsEvent: dnsEvent, startTime } = eventInterface.initializeEvent('dns', original.name, 'dns');
         if (!callback) {
             return handleFunctionWithoutCallback(
-                original, startTime, event, [address, port, callback]
+                original, startTime, dnsEvent, [address, port, callback]
             );
         }
         try {
-            eventInterface.addToMetadata(event, { address, port });
+            eventInterface.addToMetadata(dnsEvent, { address, port });
             const responsePromise = new Promise((resolve) => {
                 patchedCallback = (err, hostname, service) => {
-                    finalizeEvent(event, startTime, err, { hostname, service });
+                    eventInterface.finalizeEvent(dnsEvent, startTime, err, { hostname, service });
                     resolve();
                     if (callback) {
                         callback(err, hostname, service);
@@ -186,7 +186,7 @@ function wrapDnsLookupServiceFunction(original) {
                 };
             });
             clientRequest = original.apply(this, [address, port, patchedCallback]);
-            tracer.addEvent(event, responsePromise);
+            tracer.addEvent(dnsEvent, responsePromise);
         } catch (err) {
             tracer.addException(err);
         }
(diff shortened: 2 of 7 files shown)


src/helpers/events.js
```diff
@@ -1,4 +1,58 @@
...
+
+/**
+ * Create and initialize a new serverless event in the epsagon format.
+ * @param {string} resourceType resourceType name
+ * @param {string} name Event name
+ * @param {string} operation Operation name
+ * @param {string} origin Origin name (optional)
+ * @returns {Object} Object with dnsEvent and event start time.
+ */
+const initialEvent = (resourceType, name, operation, origin) => {
```
ranrib: initialize

src/events/dns.js
```diff
@@ -1,12 +1,9 @@
-const uuid4 = require('uuid4');
 const dns = require('dns');
 const shimmer = require('shimmer');
 const utils = require('../utils.js');
 const tracer = require('../tracer.js');
-const serverlessEvent = require('../proto/event_pb.js');
 const eventInterface = require('../event.js');
-const errorCode = require('../proto/error_code_pb.js');
-const { isBlacklistURL } = require('.././helpers/events');
+const { isBlacklistURL, initializeEvent } = require('.././helpers/events');
```
galbash: initializeEvent should be under `eventInterface`

src/events/dns.js
```diff
@@ -170,10 +143,10 @@ function wrapDnsResolveFunction(original) {
         if (typeof arg2 === 'object') {
             options = arg2;
         }
-        const { dnsEvent, startTime } = initialDnsEvent(original.name);
+        const { event, startTime } = initializeEvent('dns', original.name, 'dns');
```
galbash: I would still use `event: dnsEvent` - the word event is too generic and reading the code might become confusing

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
     };
 }
 
+const getLastSplittedString = (array) => {
```
galbash: getLastSplittedItem (doesn't have to be string). also get the delimiter as a param and maybe move to helpers

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+};
+
+/**
+ * Wrap pubsub request function.
+ * @param {Function} original pubsub request function.
+ * @returns {Function} The wrapped function
+ */
+function wrapPubSubRequestFunction(original) {
+    return function internalPubSubRequestFunction(config, callback) {
+        if (!config || !config.method) {
```
galbash: can this happen? we are not even catching a faliure here

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+            const pubsubProjectId = this.projectId;
+            const { event, startTime } = initializeEvent(
+                GOOGLE_CLOUD_TYPES.pubsub.name, pubsubProjectId, config.method
+            );
+            const requestFunctionThis = this;
+            const responsePromise = new Promise((resolve) => {
+                patchedCallback = (err, arg2, arg3) => {
+                    if ((!pubsubProjectId || pubsubProjectId === GOOGLE_CLOUD_TYPES
+                        .defaultProjectId) && !!requestFunctionThis.projectId) {
+                        event.getResource().setName(requestFunctionThis.projectId);
```
galbash: I think otherwise you sould set it to the default (or it might be empty?)

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                    }
+                    const callbackResponse = {};
+                    if (arg2) {
+                        if (config.method === 'publish') {
```
galbash: use `switch`

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                            const messageIds = arg2.messageIds ? arg2.messageIds : arg2;
+                            if (Array.isArray(messageIds) && messageIds.length) {
+                                callbackResponse.messageIds = messageIds;
+                                const [messageId] = messageIds;
```
galbash: this is not rigid enough. Either find a better way or always report messageIds

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                                callbackResponse.messageId = messageId;
+                            }
+                        } else if (config.method === 'createSubscription') {
+                            callbackResponse.subscription = arg2;
+                        } else if (config.method === 'deleteSubscription') {
+                            if (config.reqOpts && config.reqOpts.subscription) {
+                                callbackResponse.subscription = getLastSplittedString(
```
galbash: what are you splitting here? is it really crucial?

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                                    config.reqOpts.subscription
+                                );
+                            }
+                        } else if (config.method === 'deleteTopic') {
+                            if (config.reqOpts && config.reqOpts.topic) {
+                                callbackResponse.topic = getLastSplittedString(
```
galbash: same

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                            }
+                        } else if (config.method === 'createTopic') {
+                            callbackResponse.topic = arg2;
+                        }
+                    }
+                    finalizeEvent(event, startTime, err, callbackResponse);
+                    resolve();
+                    if (callback) {
+                        if (arg3) {
+                            callback(err, arg2, ...arg3);
```
galbash: you are invoking callback invalidly here. either get `(err, arg2, ...arg3)` as parameters, or use `(err, arg2, arg3)` to invoke (based on the real callback arguments).

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                                );
+                            }
+                        } else if (config.method === 'createTopic') {
+                            callbackResponse.topic = arg2;
+                        }
+                    }
+                    finalizeEvent(event, startTime, err, callbackResponse);
+                    resolve();
+                    if (callback) {
+                        if (arg3) {
```
galbash: this if is redundant. if arg3 is undefined just pass it as undefined.

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                            callback(err, arg2);
+                        }
+                    }
+                };
+            });
+            tracer.addEvent(event, responsePromise);
+        } catch (err) {
+            tracer.addException(err);
+        }
+        return original.apply(this, [config, patchedCallback]);
```
galbash, line 156: are you sure this API has no option to not pass a callback and return a promise instead? please verify that.

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+ * Wrap pubsub request function.
+ * @param {Function} original pubsub request function.
+ * @returns {Function} The wrapped function
+ */
+function wrapPubSubPullFunction(original) {
+    return function internalPubSubPullFunction(request, options, callback) {
+        try {
+            let pubsubProjectId;
+            if (request && request.subscription) {
+                const subscriptionSplited = request.subscription.split('/');
```
galbash, line 171: you have a utility function for this

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                if (subscriptionSplited.length > 1) {
+                    [, pubsubProjectId] = subscriptionSplited;
+                }
+            }
+            const { event, startTime } = initializeEvent(
+                GOOGLE_CLOUD_TYPES.pubsub.name, pubsubProjectId, 'Pull'
+            );
+            const patchedCallback = (err, res) => {
```
galbash: you are not calling the original callback!

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                        return acc;
+                    }, []);
+                    callbackResponse.messageIds = messageIds;
+                }
+                finalizeEvent(event, startTime, err, callbackResponse);
+            };
+            let responsePromise;
+            // in case callback has given from the client.
+            if (callback) {
+                responsePromise = original.apply(this, [request, options, patchedCallback]);
```
galbash: no need for a return in this block?

src/events/google_cloud.js
```diff
@@ -79,6 +87,132 @@ function bigQueryWrapper(wrappedFunction) {
...
+                patchedCallback(null, response);
+                return res;
+            }, (err) => {
+                patchedCallback(err, null);
+                throw err;
+            });
+        } catch (err) {
+            tracer.addException(err);
+        }
+        return original.apply(this, [request, options, callback]);
```
galbash: when are you reaching this line? I feel like you might be calling the original function at least twice

src/events/dns.js
galbash: I think you are not importing finalizeEvent

galbash (changes requested):
Some comments

src/events/google_cloud.js
idonava: I'm splitting the subscription name, to delete a subscription, the user use: `pubsub.subscription('subs').delete();`
But the subscription name here is "projects/pubsub-projectid/subscriptions/subs"

src/events/google_cloud.js
idonava: yes, to get the last item. here I want to get the second item.

src/events/google_cloud.js
idonava: yes, fixed

src/events/google_cloud.js
idonava: you right, fixed.

src/events/dns.js
idonava: moved.

src/events/dns.js
idonava: changed to slsEvent: dnsEvent -> soon will be push.

src/event.js
```diff
@@ -103,3 +105,51 @@ module.exports.addLabelToMetadata = function addLabelToMetadata(event, key, valu
...
+    const resource = new serverlessEvent.Resource([
+        name,
+        resourceType,
+        operation,
+    ]);
+    const slsEvent = new serverlessEvent.Event([
+        `${resourceType}-${uuid4()}`,
+        utils.createTimestampFromTime(startTime),
+        null,
+        origin || resourceType,
```
galbash: why?

src/events/google_cloud.js
```diff
@@ -79,6 +86,135 @@ function bigQueryWrapper(wrappedFunction) {
...
+                GOOGLE_CLOUD_TYPES.pubsub.name, pubsubProjectId, config.method
+            );
+            const requestFunctionThis = this;
+            const responsePromise = new Promise((resolve) => {
+                patchedCallback = (err, arg2, ...arg3) => {
+                    if ((!pubsubProjectId || pubsubProjectId === GOOGLE_CLOUD_TYPES
+                        .defaultProjectId) && !!requestFunctionThis.projectId) {
+                        pubsubEvent.getResource().setName(requestFunctionThis.projectId);
+                    }
+                    const callbackResponse = {};
```
galbash: call this `responseMetadata` instead

src/events/google_cloud.js
```diff
@@ -79,6 +86,135 @@ function bigQueryWrapper(wrappedFunction) {
...
+                    callback(err, res);
+                }
+            };
+            // in case callback has given from the client.
+            if (callback) {
+                let patchedCallbackWithPromise = callback;
+                const promise = new Promise((resolve) => {
+                    patchedCallbackWithPromise = (err, res) => {
+                        patchedCallback(err, res);
+                        resolve();
```
galbash: resolve should be called before the user callback.

src/events/google_cloud.js
```diff
@@ -79,6 +86,135 @@ function bigQueryWrapper(wrappedFunction) {
...
+                patchedCallback(null, response);
+                return res;
+            }, (err) => {
+                patchedCallback(err, null);
+                throw err;
+            });
+        } catch (err) {
+            tracer.addException(err);
+        }
+        return original.apply(this, [request, options, callback]);
```
galbash: this flow is still buggy:
1. user reaches line 201
2. user reaches line 202 - exception is raised
3. user reaches line 212
4. user reaches line 214

you called the original twice

galbash (changes requested):
still a few comments

src/event.js
idonava: my mistake of understanding the Event flow, fixed.

src/events/google_cloud.js
idonava: replaced.

src/events/google_cloud.js
idonava: tnx for that.

src/events/google_cloud.js
idonava: tnx! fixed.

src/events/google_cloud.js
```diff
@@ -79,6 +86,142 @@ function bigQueryWrapper(wrappedFunction) {
...
+                        return acc;
+                    }, []);
+                    responseMetadata.receivedMessages = receivedMessages;
+                }
+                eventInterface.finalizeEvent(pubsubEvent, startTime, err, responseMetadata);
+                if (callback) {
+                    callback(err, res);
+                }
+            };
+            // in case callback has given from the client.
```
galbash: has been given / was given

src/events/google_cloud.js
```diff
@@ -79,6 +86,142 @@ function bigQueryWrapper(wrappedFunction) {
...
+                    patchedCallbackWithPromise = (err, res) => {
+                        resolve();
+                        patchedCallback(err, res);
+                    };
+                });
+                tracer.addEvent(pubsubEvent, promise);
+                return original.apply(this, [request, options, patchedCallbackWithPromise]);
+            }
+            clientPromiseRequest = original.apply(this, [request, options, callback]);
+            tracer.addEvent(pubsubEvent, clientPromiseRequest);
```
galbash, line 222: you should add the promise after you did then on it here. do:
```
clientPromiseRequest = clientPromiseRequest.then(...)
tracer.addEvent(event, clientPromiseRequest)
```

src/events/google_cloud.js
```diff
@@ -79,6 +86,142 @@ function bigQueryWrapper(wrappedFunction) {
...
+                patchedCallback(null, response);
+                return res;
+            }, (err) => {
+                patchedCallback(err, null);
+                throw err;
+            });
+        } catch (err) {
+            tracer.addException(err);
+        }
+        if (!clientPromiseRequest) {
```
galbash: I think you should put this code in the catch (that's the only way it can run), will be more readable

galbash (changes requested):
few small comments

src/events/google_cloud.js
idonava: If tracer.addEvent throwing an error, and clientPromiseRequest was called - I don't want to call the original function twice.

src/events/google_cloud.js
```diff
@@ -79,6 +86,143 @@ function bigQueryWrapper(wrappedFunction) {
...
+                if (callback) {
+                    callback(err, res);
+                }
+            };
+            // in case callback was given from the client.
+            if (callback) {
+                let patchedCallbackWithPromise = callback;
+                const promise = new Promise((resolve) => {
+                    patchedCallbackWithPromise = (err, res) => {
+                        resolve();
```
galbash: not good enough, cause it should be *after* we update the event. you may need to split `patchedCallback`.

galbash (dismissed):
1 last comment :)

src/events/google_cloud.js
idonava: fixed.

henperez (approved):

PR #210, opened 2020-01-12
fix(consts.js): change default to https

razlo:

PR #212, opened 2020-01-14
Revert "fix(consts.js): change default to https"

razlo:
Reverts epsagon/epsagon-node#210

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -26,7 +26,7 @@ const config = {
     token: process.env.EPSAGON_TOKEN || '',
     appName: process.env.EPSAGON_APP_NAME || 'Application',
     metadataOnly: (process.env.EPSAGON_METADATA || '').toUpperCase() === 'TRUE',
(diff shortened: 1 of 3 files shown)


galbash (dismissed):

src/consts.js
```diff
@@ -10,7 +10,7 @@ if (REGION === undefined) {
 
 module.exports.REGION = REGION;
 
-module.exports.TRACE_COLLECTOR_URL = `https://${REGION}.tc.epsagon.com`;
+module.exports.TRACE_COLLECTOR_URL = `http://${REGION}.tc.epsagon.com`;
```
galbash: can't we make the default https and make the if in the previous file check `(configData.useSSL === false)`?

src/config.js
```diff
@@ -104,9 +108,9 @@ module.exports.setConfig = function setConfig(configData) {
     }
 
     // Use SSL
-    if (configData.disableSSL) {
+    if (configData.useSSL === false) {
```
galbash, line 114: change to `(configData.useSSL || configData.useSSL === false)` we want to catch true as well

galbash (approved):

PR #217, opened 2020-01-21
feat(nats): support nats publish

idonava:


src/events/nats.js
```diff
@@ -0,0 +1,170 @@
+/* eslint-disable camelcase */
```
ranrib, line 2: Disable it only where needed

idonava: If I remove this line, I need to add it again 27 times in the entire file.
I want to keep the same coding style (not camel case) of nats variables for consistency.

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+const moduleUtils = require('./module_utils.js');
+
+const NATS_TYPES = {
+    name: 'nats',
+    mainWrappedFunction: 'Client',
+    badMessage: 'NATS_BAD_JSON_MSG',
+};
+
+const getServerHostname = (currentServer) => {
+    let serverHostname = NATS_TYPES.serverDefaultHostname;
```
galbash: you can use the trinary operator ( `a ? b : c`) here instead

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+    if (currentServer.url && currentServer.url.hostname) {
+        serverHostname = currentServer.url.hostname;
+    }
+    return serverHostname;
+};
+
+const getPublishParams = (subject, msg, opt_reply, opt_callback, jsonConnectProperty) => {
```
galbash, line 34: docs?

galbash, line 34: is this logic copied from nuts? (the best answer would be yes)

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+                            null,
+                            responseMetadata
+                        );
+                        resolve();
+                        opt_callback_internal();
+                    };
+                });
+                tracer.addEvent(natsPublishEvent, promise);
+            } else {
+                tracer.addEvent(natsPublishEvent);
```
galbash: reverse the order of operations in this if. also, I see why you don't want to add a callback if one doesn't exist, but let's talk about it.

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+                eventInterface.finalizeEvent(
+                    natsPublishEvent,
+                    startTime,
+                    null,
+                    responseMetadata
+                );
+            }
+            return original.apply(this,
+                [subject_internal, msg_internal, opt_reply_internal, patchedCallback]);
```
galbash: this seems risky. why not use the original params?

galbash: just remove this call to outside of the try / catch. currently, you are never assigning clientRequest and if the original function throws an error you call it twice!!

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+        const connectFunctionResponse = connectFunction(url, opts);
+        if (connectFunctionResponse && connectFunctionResponse.constructor) {
+            if (connectFunctionResponse.constructor.name !== NATS_TYPES.mainWrappedFunction) {
+                return connectFunctionResponse;
+            }
+            const serverHostname = getServerHostname(connectFunctionResponse.currentServer);
+            const jsonConnectProperty = connectFunctionResponse.options ?
+                connectFunctionResponse.options.json : null;
+            shimmer.wrap(connectFunctionResponse, 'publish', () => wrapNatsPublishFunction(connectFunctionResponse.publish, serverHostname, jsonConnectProperty));
+        }
```
galbash: maybe wrap all of this with a try/catch block

galbash (dismissed):
looks good, one major issue and few minor comments

src/events/nats.js
idonava: I think its more readable and clearer instead: 
`const getServerHostname = (currentServer) => (currentServer.url && currentServer.url.hostname) ? currentServer.url.hostname : NATS_TYPES.serverDefaultHostname
`

src/events/nats.js
idonava: added.

idonava: yes

src/events/nats.js
idonava: Because I already build the right params. But maybe you right and its better to pass the original params, in case of nats change is api.

idonava: instead of return this original.apply, I assign it as ClientRequest, and changed the three first params to the original params.

src/events/nats.js
idonava: ok

src/events/nats.js
idonava: changed.

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
+/* eslint-disable no-console */
```
galbash, line 1: remove this line

henperez (approved):

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Wrap nats connect function.
+ * @param {Function} connectFunction nats connect function.
+ * @returns {Function} The wrapped function
+ */
+function wrapNatsConnectFunction(connectFunction) {
+    return function internalNatsConnectFunction(url, opts) {
+        let connectFunctionResponse;
+        try {
+            connectFunctionResponse = connectFunction(url, opts);
```
galbash, line 141: Use apply

src/events/nats.js
```diff
@@ -0,0 +1,170 @@
...
+                if (connectFunctionResponse.constructor.name === NATS_TYPES.mainWrappedFunction) {
+                    const serverHostname = getServerHostname(connectFunctionResponse.currentServer);
+                    const jsonConnectProperty = connectFunctionResponse.options ?
+                        connectFunctionResponse.options.json : null;
+                    shimmer.wrap(connectFunctionResponse, 'publish', () => wrapNatsPublishFunction(connectFunctionResponse.publish, serverHostname, jsonConnectProperty));
+                }
+            }
+        } catch (err) {
+            if (!connectFunctionResponse) {
+                connectFunctionResponse = connectFunction.apply(this, [url, opts]);
```
galbash, line 152: You may be calling this twice if the original function raised an error

src/events/nats.js
idonava: removed.

src/events/nats.js
idonava: changed.

src/events/nats.js
idonava: fixed

PR #222, opened 2020-01-27
openwhisk: various fixes

### LINKED ISSUE #221: openwhisk:  invoking sub-actions causes process to hang on errors
if an openhwisk action invokes another actions, and that action responds with an error, the process seems to hang.

### LINKED ISSUE #223: openwhisk: actions invoked with 'result:true' don't show result in traces
if an openwhisk action is invoked with `result:true`, the API directly returns the result of the action and not the entire response object. in this case, the trace doesn't contain information about the return value.

### LINKED ISSUE #224: openwhisk: params of action invocations not recorded in traces
it would be practical to see the params of action invocations in the traces. currently they are not added to the metadata map.

tripodsan:
- fixes #221

The problem was, that the tracer that receives the `responsePromise` would never continue, because the `responsePromise` would never resolve or reject in case of an error by the `request` promise.

- fixes #223 

actions invoked with `result:true` were not returning the entire response, so the response was not properly recorded.

- fixes #224 

also added the params of the actions to be invoked

### DIFF
--- a/src/events/openwhisk.js
+++ b/src/events/openwhisk.js
@@ -31,6 +31,9 @@ function openWhiskWrapper(wrappedFunction) {
         ]);
 
         invokeEvent.setResource(resource);
+        eventInterface.addToMetadata(invokeEvent, {
+            params: options.params,
+        });
         let request;
         let response;
         if (options.result) {
--- a/src/events/openwhisk.js
+++ b/src/events/openwhisk.js
@@ -33,7 +33,7 @@ function openWhiskWrapper(wrappedFunction) {
         invokeEvent.setResource(resource);
 
         const request = wrappedFunction.apply(this, [options, callback]);
-        const responsePromise = new Promise((resolve) => {
+        const responsePromise = new Promise((resolve, reject) => {
             request.then((res) => {
                 eventInterface.addToMetadata(
                     invokeEvent,
@@ -44,6 +44,9 @@ function openWhiskWrapper(wrappedFunction) {
                 );
                 invokeEvent.setDuration(utils.createDurationTimestamp(startTime));
                 resolve();
+            }).catch((err) => {
+                eventInterface.setException(invokeEvent, err);
+                reject(err);
             });
         });
 
--- a/src/events/openwhisk.js
+++ b/src/events/openwhisk.js
@@ -31,15 +31,37 @@ function openWhiskWrapper(wrappedFunction) {
         ]);
 
         invokeEvent.setResource(resource);
+        let request;
+        let response;
+        if (options.result) {
+            // action.invoke would return directly `response.result` so we would loose some
+            // of the information from the response.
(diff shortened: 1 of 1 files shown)


src/events/openwhisk.js
```diff
@@ -31,24 +31,52 @@ function openWhiskWrapper(wrappedFunction) {
...
                         activation_id: res.activationId,
-                        response: res.response,
+                        response: resp,
                     }
                 );
                 invokeEvent.setDuration(utils.createDurationTimestamp(startTime));
                 resolve();
+            }).catch((err) => {
+                eventInterface.setException(invokeEvent, err);
+                reject(err);
```
galbash: let's resolve here (we are deferring it anyway, and this promise should generally be resolved when the request is done). We can actually use finally here and put the resolve there (instead of in the then and catch clauses). What do you think?

src/events/openwhisk.js
```diff
@@ -31,24 +31,52 @@ function openWhiskWrapper(wrappedFunction) {
         ]);
 
         invokeEvent.setResource(resource);
+        eventInterface.addToMetadata(invokeEvent, {
+            params: options.params,
```
galbash, line 35: maybe let's pass this as the third argument? (so it won't be recorded when metadataOnly=True)

src/events/openwhisk.js
```diff
@@ -31,24 +31,52 @@ function openWhiskWrapper(wrappedFunction) {
...
+                let resp = res.response;
+                if (resp && resp.result && resp.result.body && resp.result.body.length > 100) {
+                    // create copy so we can trim the long response body
+                    resp = Object.assign({}, resp);
+                    resp.result = Object.assign({}, resp.result);
+                    resp.result.body = `${resp.result.body.substring(0, 100)}...(truncated)`;
+                }
                 eventInterface.addToMetadata(
                     invokeEvent,
                     {
```
galbash, line 71: let's move this to metadataOnly=false as well (use the third argument instead of the second)

galbash (commented):
looks good, @ranrib can you have a look at the comments to fit it to our conventions and merge it?

src/events/openwhisk.js
tripodsan: but then it is not consistent with the data that is added in the runner. there all the params are added to the default map:
https://github.com/epsagon/epsagon-node/blob/abcf003c2ab88170e018f9aba650d537d1cb6332/src/wrappers/openwhisk.js#L73-L79

src/events/openwhisk.js
tripodsan: ok.

src/events/openwhisk.js
tripodsan: ok, I exclude the body and the headers from the default map....

ranrib (approved):

PR #225, opened 2020-01-28
feat(config): adding sendTimeout option to config

galbash:
This PR adds the `sendTimeout` config option. now the send timeout of the tracer can be configured via `epsagon.init` and not just via an environment variable

### DIFF
--- a/test/unit_tests/test_config.js
+++ b/test/unit_tests/test_config.js
@@ -81,5 +81,16 @@ describe('tracer config tests', () => {
         const sendTimeout = 1000;
         config.setConfig({ sendTimeout });
         expect(config.getConfig().sendTimeout).to.be.equal(sendTimeout);
+
+        const sendTimeoutString = '1000';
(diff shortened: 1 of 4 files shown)


test/unit_tests/test_config.js
```diff
@@ -75,4 +76,10 @@ describe('tracer config tests', () => {
         config.setConfig({ httpErrorStatusCode });
         expect(config.HTTP_ERR_CODE).to.be.equal(httpErrorStatusCode);
     });
+
+    it('setConfig: set custom sendTimeout', () => {
+        const sendTimeout = 1000;
+        config.setConfig({ sendTimeout });
+        expect(config.getConfig().sendTimeout).to.be.equal(sendTimeout);
+    });
```
haimrait, line 95: Add test for case that sendTimeout is string

haimrait (dismissed):

haimrait (approved):

PR #230, opened 2020-02-05
feat(lambda): adding support for non-error 1st callback arg

galbash:


ranrib (dismissed):

src/wrappers/lambda.js
```diff
@@ -97,7 +97,21 @@ function baseLambdaWrapper(
...
             if (error) {
                 // not catching false here, but that seems OK
-                eventInterface.setException(runner, error, false);
+                let reportedError = error;
+                if (!error.name) {
+                    let errorMessage;
+                    try {
+                        errorMessage = typeof error === 'string' ? error : JSON.stringify(error);
+                    } catch (stringifyErr) {
+                        errorMessage = 'Unserializable Error';
```
adavidai: Avoid overriding the user error with an error in the instrumentation itself

test/unit_tests/wrappers/test_lambda.js
```diff
@@ -258,6 +258,25 @@ describe('lambdaWrapper tests', () => {
         }, 1);
     });
 
+    it('lambdaWrapper: wrapped callback string as error', (done) => {
```
adavidai, line 261: Consider enhancing the tests to include other object types for error

adavidai (commented):
Looks good 👍 
See some comments below

yotamelkaslasy (approved):

PR #232, opened 2020-02-06
Feature/support sqs in stepfunctions/ep 2317

idonava:
image

### DIFF
--- a/src/events/aws_sdk.js
+++ b/src/events/aws_sdk.js
@@ -606,6 +606,7 @@ const stepFunctionsEventCreator = {
                 'Start Date': `${response.data.startDate}`,
             });
             break;
+
         default:
             break;
         }
@@ -693,8 +694,6 @@ function AWSSDKWrapper(wrappedFunction) {
             const { serviceIdentifier } = (
                 request.service.constructor.prototype
             );
-            // eslint-disable-next-line no-console
-            console.log('serviceIdentifier', serviceIdentifier);
             if (!(serviceIdentifier in specificEventCreators)) {
                 // resource is not supported yet
                 return wrappedFunction.apply(this, [callback]);
--- a/src/events/aws_sdk.js
+++ b/src/events/aws_sdk.js
@@ -515,6 +515,33 @@ const athenaEventCreator = {
     },
 };
 
+/**
+ * Generate a new stpes_dict id, set the step number as -1 to mark the invoker,
+ * and adding the new steps_dict to the event metadata.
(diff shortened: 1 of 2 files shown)


src/events/aws_sdk.js
```diff
@@ -546,6 +546,24 @@ const stepFunctionsEventCreator = {
 
             break;
         }
+        case 'sendTaskSuccess': {
```
ronyis, line 557: Missing `sendTaskFailure` (?)

src/events/aws_sdk.js
```diff
@@ -546,6 +546,24 @@ const stepFunctionsEventCreator = {
...
+            let output;
```
ronyis: This code is a duplication of handling `startExecution`

src/triggers/aws_lambda.js
```diff
@@ -115,6 +115,13 @@ function createSQSTrigger(event, trigger) {
     }, {
         'Message Body': event.Records[0].body,
     });
+    const messageBody = JSON.parse(event.Records[0].body || {});
```
ronyis: A comment is really needed (explain that you are extracting the data for the case for SQS task in step functions and so on)

ronyis (dismissed):

ronyis:
We should better add a test for that

ronyis:
What about the issue with resource name not set?

src/events/aws_sdk.js
idonava: I didn't see a reason to catch sendTaskFailure, Because it's the last step of the machine, And I don't want to pass it with a new id and step_num=-1. 
image

src/events/aws_sdk.js
idonava: fixed

src/triggers/aws_lambda.js
idonava: added.

src/events/aws_sdk.js
ronyis: I think that it doesn't always have to be the last step (you can configure task transition after failure), but anyway I see that there isn't an `output` field to use

ronyis (commented):
Code looks 💯 
But I think we should add basic tests mostly to make sure that we don't change the patched data.
@henprtz can decide on that.

henperez:
Thx @ronyis - I agree this requires us to add basic UT

henperez (approved):

PR #233, opened 2020-02-09
fix(pubsub): add message body to publish

idonava:
Adding a message body to pubsub publish and fix instrumentation.


ronyis (commented):
Looks good to me, I think @galbash should look

galbash (approved):

PR #234, opened 2020-02-15
feat(http2.js): add support for HTTP2

ranrib:


src/events/http2.js
```diff
@@ -0,0 +1,166 @@
+/**
+ * @fileoverview Handlers for http & https libraries instrumentation
+ */
+
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const urlLib = require('url');
+const http2 = require('http2');
```
galbash: maybe should be tryRequire (not supported in older versions)

src/events/http2.js
```diff
@@ -0,0 +1,166 @@
...
+const utils = require('../utils.js');
+const tracer = require('../tracer.js');
+const serverlessEvent = require('../proto/event_pb.js');
+const eventInterface = require('../event.js');
+const errorCode = require('../proto/error_code_pb.js');
+const { isBlacklistURL, isBlacklistHeader } = require('.././helpers/events');
+const {
```
galbash, line 11: might be better to patch both in the same file? how different are they?

src/events/http2.js
ranrib: HTTP2 appears since node8. I think we will just remove the support for node6 because that's very old

src/events/http2.js
ranrib: The tracing has some very significant difference both in the patching and in the wrapping, so I wouldn't recommend it.
I did export common HTTP functions from http.js

src/events/http2.js
ranrib: Fixed

src/events/http.js
```diff
@@ -53,6 +53,37 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+
+
+/**
+ * Checks if API Gateway details appear in the headers, and update event accordingly
+ * @param {object} headers data
+ * @param {Resource} resource object
+ * @param {Event} httpEvent object
+ */
+function updateAPIGateway(headers, resource, httpEvent) {
+    if ('x-amzn-requestid' in headers) {
```
galbash: check `headers` is not null/undefined

src/events/http.js
```diff
@@ -392,4 +412,10 @@ module.exports = {
             WreckWrapper
         );
     },
+    isURLIgnoredByUser,
+    resolveHttpPromise,
+    USER_AGENTS_BLACKLIST,
+    URL_BLACKLIST,
+    generateEpsagonTraceId,
+    updateAPIGateway,
```
galbash: move to a common location? not a must let's talk about this

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
+/**
+ * @fileoverview Handlers for http & https libraries instrumentation
```
galbash: fix

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+ */
+function extractHeaders(headers) {
+    return Object.entries(headers)
+        .filter(header => !header[0].startsWith(':'))
+        .reduce((obj, header) => {
+            const [key, value] = header;
+            obj[key] = value; // eslint-disable-line no-param-reassign
+            return obj;
+        }, {});
+}
```
galbash: explain this to me

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+            }).catch((err) => {
+                tracer.addException(err);
+            });
+
+            tracer.addEvent(httpEvent, responsePromise);
+        } catch (error) {
+            tracer.addException(error);
+        }
+
+        if (!clientRequest) {
```
galbash: If the original function throws an error, you call it twice

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
+/**
+ * @fileoverview Handlers for http & https libraries instrumentation
+ */
+
+const uuid4 = require('uuid4');
+const shimmer = require('shimmer');
+const urlLib = require('url');
```
galbash, line 6: is this a core library? is it available in all versions or should you tryRequire it as well?

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+const utils = require('../utils.js');
+const tracer = require('../tracer.js');
+const serverlessEvent = require('../proto/event_pb.js');
+const eventInterface = require('../event.js');
+const errorCode = require('../proto/error_code_pb.js');
+const { isBlacklistURL, isBlacklistHeader } = require('.././helpers/events');
```
galbash: the `/./` is redundant

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+ */
+function httpWrapper(wrappedFunction, authority) {
+    return function internalHttpWrapper(headers, options) {
+        let clientRequest = null;
+        try {
+            const { hostname } = urlLib.parse(authority);
+
+            const reqHeaders = extractHeaders(headers);
+
+            if (isBlacklistURL(hostname, URL_BLACKLIST, headers[':path']) || isURLIgnoredByUser(hostname)) {
```
galbash, line 56: why are you checking `headers` if you just extracted `reqHeaders`? 
anyway these 2 tests should come before parsing the hostname and extracting the headers

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+                utils.debugLog(`filtered blacklist hostname ${hostname}`);
+                return wrappedFunction.apply(this, [headers, options]);
+            }
+            if (isBlacklistHeader(reqHeaders, USER_AGENTS_BLACKLIST)) {
+                utils.debugLog(`filtered blacklist headers ${JSON.stringify(reqHeaders)}`);
```
galbash: calling JSON.stringify here will hurt performance

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            // Inject header to support tracing over HTTP requests to opentracing monitored code
+            const epsagonTraceId = generateEpsagonTraceId();
+            headers['epsagon-trace-id'] = epsagonTraceId; // eslint-disable-line no-param-reassign
+
+            const resource = new serverlessEvent.Resource([
+                hostname,
+                'http',
+                headers[':method'],
```
galbash, line 74: why not `reqHeaders`?

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+            if (isBlacklistHeader(reqHeaders, USER_AGENTS_BLACKLIST)) {
+                utils.debugLog(`filtered blacklist headers ${JSON.stringify(reqHeaders)}`);
+                return wrappedFunction.apply(this, [headers, options]);
+            }
+
+            // Inject header to support tracing over HTTP requests to opentracing monitored code
+            const epsagonTraceId = generateEpsagonTraceId();
+            headers['epsagon-trace-id'] = epsagonTraceId; // eslint-disable-line no-param-reassign
+
+            const resource = new serverlessEvent.Resource([
```
galbash: use @idonava 's new event interface methods for initializing and finalizing an event

src/events/http2.js
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Wraps the request method of HTTP2 connect
+ * @param {Function} connectFunction connect function
+ * @return {Function} the wrapped function
+ */
+function wrapHttp2Connect(connectFunction) {
+    return function innerWrapHttp2Connect(authority, options, listener) {
+        const clientSession = connectFunction.apply(this, [authority, options, listener]);
+        shimmer.wrap(clientSession, 'request', wrappedFunction => httpWrapper(wrappedFunction, authority));
```
galbash: wrap this with a try / catch

galbash (changes requested):
few comments

src/events/http.js

src/events/http.js

src/events/http2.js
ranrib: 👍

src/events/http2.js
ranrib: Added more detailed docs.

src/events/http2.js
ranrib: Yes, this is from version 4.x. It's also being used in the original http module.

src/events/http2.js
ranrib: 👍

src/events/http2.js
ranrib: For these 2 tests we need - path, hostname, and headers ("cleaned" ones). I'm not sure what is the problem here.

src/events/http2.js
ranrib: 👍

src/events/http2.js
ranrib: This is the method. reqHeaders include only real HTTP headers

src/events/http2.js
ranrib: 👍

src/events/http2.js
ranrib: 👍

src/events/http2.js
ranrib: 👍

ranrib:
@galbash thanks - please re-review

src/event.js
```diff
@@ -130,7 +130,7 @@ module.exports.initializeEvent = function initializeEvent(resourceType, name, op
         errorCode.ErrorCode.OK,
     ]);
     slsEvent.setResource(resource);
-    return { slsEvent, startTime };
+    return { slsEvent, startTime, resource };
```
galbash: This is redundent (you can get the resource from the event)

src/events/http2.js
```diff
@@ -0,0 +1,169 @@
...
+                    path: headers[':path'],
+                    request_headers: reqHeaders,
+                });
+
+            try {
+                clientRequest = wrappedFunction.apply(this, [headers, options]);
+            } catch (err) {
+                eventInterface.setException(httpEvent, err);
+                tracer.addEvent(httpEvent);
+                throw err;
```
galbash: This will be caught by the external try/catch and not be propagated to the user.

src/events/http2.js
```diff
@@ -0,0 +1,169 @@
...
+ * @param {string} authority hostname
+ * @returns {Function} The wrapped function
+ */
+function httpWrapper(wrappedFunction, authority) {
+    return function internalHttpWrapper(headers, options) {
+        let clientRequest = null;
+        try {
+            const { hostname } = urlLib.parse(authority);
+
+            const reqHeaders = extractHeaders(headers);
```
galbash: Should be blow the next 2 ifs (the urllib parse as well)

src/events/http2.js
```diff
@@ -0,0 +1,169 @@
...
+                clientRequest.once('response', (res) => {
+                    updateAPIGateway(res, resource, httpEvent);
+                    eventInterface.addToMetadata(httpEvent, {
+                        status: res[':status'],
+                    }, {
+                        response_headers: extractHeaders(res),
+                    });
+                });
+            }).catch((err) => {
+                tracer.addException(err);
```
galbash, line 129: No resolve here?

galbash (changes requested):
Still a few comments

src/event.js
ranrib: 👍

src/events/http2.js
ranrib: Moved it, but the hostname is needed.. can't move it below

src/events/http2.js
ranrib: Fixed

ranrib: Can't resolve in here - this is the catch block, means that the promise exploeded

ranrib:
@galbash all fixed.
I've added also the right way to extract the request body in http, since it wasn't being made right so far.

all changes tested with got, fetch, axios, request, superagent, http2, fetch-h2, Wreck.

src/events/http.js
```diff
@@ -256,15 +226,31 @@ function httpWrapper(wrappedFunction) {
...
+            function endWrapper(wrappedEndFunc) { // eslint-disable-line no-inner-declarations
+                return function internalEndWrapper(...args) {
+                    if (
+                        (!body || body === '') && args[0] && (
+                            (args[0] instanceof String) || (args[0] instanceof Buffer)
+                        )
+                    ) {
+                        setJsonPayload(httpEvent, 'request_body', args[0]);
+                    }
+                    const result = wrappedEndFunc.apply(this, args);
```
galbash: you should do this anyway (try catch the above, return this). isn't there a chance you will catch it twice?

src/events/http2.js
```diff
@@ -0,0 +1,172 @@
...
+ * @param {string} authority hostname
+ * @returns {Function} The wrapped function
+ */
+function httpWrapper(wrappedFunction, authority) {
+    return function internalHttpWrapper(headers, options) {
+        let clientRequest = null;
+        let httpEvent = null;
+        let startTime = null;
+        try {
+            const { hostname } = urlLib.parse(authority);
```
galbash, line 54: this should be below the 2 ifs

src/events/http2.js
```diff
@@ -0,0 +1,172 @@
...
+
+            if (isBlacklistURL(hostname, URL_BLACKLIST, headers[':path']) || isURLIgnoredByUser(hostname)) {
+                utils.debugLog(`filtered blacklist hostname ${hostname}`);
+                return wrappedFunction.apply(this, [headers, options]);
+            }
+
+            const reqHeaders = extractHeaders(headers);
```
galbash, line 61: this should be below the 2 ifs

src/events/http2.js
```diff
@@ -0,0 +1,172 @@
...
+                    http_trace_id: epsagonTraceId,
+                }, {
+                    path: headers[':path'],
+                    request_headers: reqHeaders,
+                });
+        } catch (error) {
+            tracer.addException(error);
+            if (httpEvent) {
+                tracer.addEvent(httpEvent);
+            }
```
galbash: I don't understand what you are doing here

src/helpers/http.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param {object} httpEvent The current event
+ * @param {string} key name in metadata
+ * @param {string} data data to jsonify
+ */
+function setJsonPayload(httpEvent, key, data) {
+    try {
+        const payload = JSON.parse(data);
+        eventInterface.addToMetadata(httpEvent, {}, {
+            [key]: payload,
+        });
```
galbash: no need to `JSON.parse` the data (actually, you shouldn't even since the proto object get a str-str map)

galbash (changes requested):
few small things

src/events/http.js

src/events/http2.js
ranrib: It still can't be - this is being used in the first if

src/events/http2.js
ranrib: This is being used by the second if

src/events/http2.js
ranrib: There might be an event here, not necessarily because the code might explode before

src/helpers/http.js
ranrib: 1. I think we should capture only JSON data and not free text
2. It will be parsed more easily in Epsagon

galbash (approved):

PR #238, opened 2020-02-27
feat(nats): propagate nats id for JSON-based msgs

idonava:
1. add epsagon to nats messages only if: 
   a. process.env.EPSAGON_PROPAGATE_NATS_ID=TRUE
   b. is a JSON based message.
2. update Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,9 @@ An environment variable must be added to start tracing the following libraries:
 
 dns: `EPSAGON_DNS_INSTRUMENTATION = true`
 
+NATS: `EPSAGON_PROPAGATE_NATS_ID = TRUE` Propagate Nats Identifier for JSON-based messages.
+
+
 ## Copyright
 
 Provided under the MIT license. See LICENSE for details.
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Support for Express, Hapi, and other frameworks is done through [epsagon-framewo
 ## Optional instrumentations
 An environment variable must be added to start tracing the following libraries:
 
-dns: `EPSAGON_DNS_INSTRUMENTATION = true`
+dns: `EPSAGON_DNS_INSTRUMENTATION = TRUE`
 
 NATS: `EPSAGON_PROPAGATE_NATS_ID = TRUE` Propagate Nats Identifier for JSON-based messages.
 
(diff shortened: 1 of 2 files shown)


ranrib (dismissed):

README.md
```diff
@@ -140,6 +140,9 @@ An environment variable must be added to start tracing the following libraries:
 
 dns: `EPSAGON_DNS_INSTRUMENTATION = true`
 
+NATS: `EPSAGON_PROPAGATE_NATS_ID = TRUE` Propagate Nats Identifier for JSON-based messages.
```
ronyis, line 143: Choose if `TRUE` or `true` for both :)

ronyis (dismissed):
One comment

README.md
idonava: fixed :)

src/events/nats.js
```diff
@@ -55,6 +56,9 @@ const getPublishParams = (subject, msg, opt_reply, opt_callback, jsonConnectProp
         opt_callback_internal = opt_reply;
         opt_reply_internal = undefined;
     }
+    if (jsonConnectProperty && (process.env.EPSAGON_PROPAGATE_NATS_ID || '').toUpperCase() === 'TRUE') {
```
ronyis, line 59: I think we would like to use `TRUE` in default. Please check with @ranrib / @henprtz

ronyis (approved):

src/events/nats.js
idonava: I talked about it with ran, He told me to use default `FALSE`

README.md
galbash: maybe disable only if set to false (reverse behaviour)

galbash: maybe disable only if set to false (reverse behavior)

src/events/nats.js
```diff
@@ -55,6 +56,9 @@ const getPublishParams = (subject, msg, opt_reply, opt_callback, jsonConnectProp
...
+        msg_internal.epsagon_id = uuid4();
```
galbash, line 60: every message in nats already has an sid, no need to inject a new one (I can point you to the docs)

galbash (changes requested):

galbash (approved):

PR #241, opened 2020-03-12
fix(http.js): fix http response decoding

ranrib:
Parsing response data in chucks, and decoding it accordingly (the way `http` handles it).
Decoding will be done by default unless `EPSAGON_DECODE_HTTP !== true`.
In addition, stop relying on request close event, and changing it to response event to terminate the Epsagon event, since it doesn't work for keepalive

### DIFF
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -277,7 +277,7 @@ function httpWrapper(wrappedFunction) {
 
 
             const responsePromise = new Promise((resolve) => {
-                let data = '';
+                const chunks = [];
 
                 let isTimeout = false;
                 clientRequest.on('timeout', () => {
@@ -305,14 +305,11 @@ function httpWrapper(wrappedFunction) {
                     }
                 });
 
-                clientRequest.on('close', () => {
-                    resolveHttpPromise(httpEvent, resolve, startTime);
-                });
-
                 clientRequest.on('response', (res) => {
-                    res.on('data', (chunk) => { data += chunk; });
+                    res.on('data', (chunk) => { chunks.push(chunk); });
                     res.on('end', () => {
-                        setJsonPayload(httpEvent, 'response_body', data);
+                        setJsonPayload(httpEvent, 'response_body', Buffer.concat(chunks), res.headers['content-encoding']);
+                        resolveHttpPromise(httpEvent, resolve, startTime);
                     });
                 });
             }).catch((err) => {
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -14,6 +14,7 @@ const eventInterface = require('../event.js');
 const errorCode = require('../proto/error_code_pb.js');
 const config = require('../config.js');
 const moduleUtils = require('./module_utils.js');
+const { MAX_HTTP_VALUE_SIZE } = require('../consts.js');
 const { isBlacklistURL, isBlacklistHeader } = require('../helpers/events');
 const {
     isURLIgnoredByUser,
@@ -306,7 +307,12 @@ function httpWrapper(wrappedFunction) {
                 });
 
                 clientRequest.on('response', (res) => {
-                    res.on('data', (chunk) => { chunks.push(chunk); });
+                    res.on('data', (chunk) => {
+                        const totalSize = chunks.reduce((total, item) => item.length + total, 0);
+                        if (totalSize + chunk.length <= MAX_HTTP_VALUE_SIZE) {
+                            chunks.push(chunk);
+                        }
+                    });
                     res.on('end', () => {
                         setJsonPayload(httpEvent, 'response_body', Buffer.concat(chunks), res.headers['content-encoding']);
                         resolveHttpPromise(httpEvent, resolve, startTime);
--- a/src/events/http2.js
+++ b/src/events/http2.js
@@ -68,7 +68,7 @@ function httpWrapper(wrappedFunction, authority) {
             const epsagonTraceId = generateEpsagonTraceId();
             headers['epsagon-trace-id'] = epsagonTraceId; // eslint-disable-line no-param-reassign
 
-            const { slsEvent, eventStartTime } = eventInterface.initializeEvent(
+            const { slsEvent, startTime: eventStartTime } = eventInterface.initializeEvent(
                 'http',
                 hostname,
                 headers[':method'],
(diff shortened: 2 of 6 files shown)


src/events/http.js
```diff
@@ -305,14 +305,11 @@ function httpWrapper(wrappedFunction) {
                     }
                 });
 
-                clientRequest.on('close', () => {
-                    resolveHttpPromise(httpEvent, resolve, startTime);
-                });
-
                 clientRequest.on('response', (res) => {
-                    res.on('data', (chunk) => { data += chunk; });
+                    res.on('data', (chunk) => { chunks.push(chunk); });
```
tripodsan: what happens if the response is too large? e.g. a 10gb movie ? (not that we have those :-)

src/events/http.js
```diff
@@ -305,14 +305,11 @@ function httpWrapper(wrappedFunction) {
...
                     res.on('end', () => {
-                        setJsonPayload(httpEvent, 'response_body', data);
+                        setJsonPayload(httpEvent, 'response_body', Buffer.concat(chunks), res.headers['content-encoding']);
```
tripodsan, line 317: what happens if the response doesn't have a `content-encoding` header?

src/events/http2.js
```diff
@@ -113,16 +114,17 @@ function httpWrapper(wrappedFunction, authority) {
                 });
 
                 clientRequest.once('close', () => {
-                    setJsonPayload(httpEvent, 'response_body', data);
+                    setJsonPayload(httpEvent, 'response_body', Buffer.concat(chunks), responseHeaders['content-encoding']);
```
tripodsan, line 123: what happens if the response doesn't have a `content-encoding` header?

src/helpers/http.js
```diff
@@ -18,6 +19,18 @@ const URL_BLACKLIST = {
...
+// Brotli decompression exists since Node v10
+const ENCODING_FUNCTIONS = {
+    br: zlib.brotliDecompressSync,
+    brotli: zlib.brotliDecompressSync,
+    gzip: zlib.gunzipSync,
+    deflate: zlib.deflateSync,
+};
+
+// Whether to try and decode HTTP response
+const DECODE_HTTP = (process.env.EPSAGON_DECODE_HTTP || 'TRUE').toUpperCase() === 'TRUE';
```
tripodsan: it would be probably good to have this as config option as well.

src/helpers/http.js
```diff
@@ -47,12 +60,21 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
  * @param {object} httpEvent The current event
  * @param {string} key name in metadata
  * @param {string} data data to jsonify
+ * @param {string} encoding data encoding from the headers
  */
-function setJsonPayload(httpEvent, key, data) {
+function setJsonPayload(httpEvent, key, data, encoding) {
```
tripodsan, line 62: shouldn't you call this method only, if the response content type is actually `application/json` ?

src/helpers/http.js
```diff
@@ -47,12 +60,21 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
     try {
-        JSON.parse(data);
+        let jsonData = data;
+        if (DECODE_HTTP && ENCODING_FUNCTIONS[encoding]) {
+            try {
+                jsonData = ENCODING_FUNCTIONS[encoding](data);
+            } catch (err) {
+                utils.debugLog(`Could not parse ${encoding} ${key} in http`);
```
tripodsan:
```suggestion
                utils.debugLog(`Could decode ${key} with ${encoding} in http`);
```

src/helpers/http.js
```diff
@@ -47,12 +60,21 @@ function resolveHttpPromise(httpEvent, resolveFunction, startTime) {
...
+            }
+        }
+        JSON.parse(jsonData);
         eventInterface.addToMetadata(httpEvent, {}, {
-            [key]: data,
+            [key]: jsonData.toString(),
```
tripodsan, line 74: why not invoke `toString()` in encoding block above?

tripodsan (dismissed):
I tested it and it works (alomst :-) :

- using helix-fetch no longer makes the actions hang
- returns the correct response using http1, and shows the trace in the epsagon dashboard
- returns the correct response using http2, but the traces never show up in the dashboard

src/events/http.js
ranrib: You're right. Before we send the trace we make sure it does not cross the limit (64KB if I recall correctly), and if so we remove such fields.

src/events/http.js
ranrib: It will be `undefined` which means that the code will not try to decode it.

src/events/http2.js
ranrib: It will be `undefined` which means that the code will not try to decode it.

src/helpers/http.js
ranrib: because in case it is a not encoded data, the `data` field is type `Buffer` which needs to be converted to string. Same applies for the decoding functions

src/helpers/http.js
ranrib: 👍

src/helpers/http.js
ranrib: In general, this is a true approach, but not all APIs are always sending the right header. For example, the GitHub API returns `'content-type': 'text/plain; charset=utf-8'` with a gzip that insides there's a JSON.

ranrib:
> * returns the correct response using http2, but the traces never show up in the dashboard

You're right - this fixes the bug: https://github.com/epsagon/epsagon-node/pull/241/files#diff-9bf9e9be548a06d15dc52a6462588c5dR71

(the duration was null, hence, the event was thrown)

src/events/http.js
galbash: I think @tripodsan is right, and we should add some limit here (first 5 chinks?) To not use too much memory

src/helpers/http.js
tripodsan: but I guess only if the `accept` request header was `application/json` ?
https://developer.github.com/v3/#current-version

src/helpers/http.js
tripodsan: but then the `JSON.parse()` above still works ?
wouldn't it be better to call `jsonData.toString('utf-8')` ?

src/events/http.js
ranrib: 👍 Done

galbash (approved):

PR #242, opened 2020-03-16
feat(mqtt): add mqtt support

haimrait:
Add support to mqtt publish function and constructor

### DIFF
--- a/src/events/mqtt.js
+++ b/src/events/mqtt.js
@@ -1,0 +1,109 @@
+const shimmer = require('shimmer');
+const tracer = require('../tracer.js');
+const eventInterface = require('../event.js');
+const moduleUtils = require('./module_utils.js');
+
+
+const getPublishParams = (topic, message, options, callback) => {
+    const topic_internal = topic;
+    const message_internal = message;
+    const options_internal = options;
+    const callback_internal = callback;
+
+    return {
+        topic_internal,
+        message_internal,
+        options_internal,
+        callback_internal
+    };
+};
+
+function publishWrapper(wrappedFunction) {
+    return function internalPublishWrapper(topic, message, options, callback) {
+        const {
+                topic_internal,
+                message_internal,
+                options_internal,
+                callback_internal
+            } = getPublishParams(topic, message, options, callback);
+        let patchedCallback = callback_internal;   
+
+        try {
+                const { slsEvent: mqttEvent, startTime } = eventInterface.initializeEvent('MQTT',
+                                                                                          this.options.host,
+                                                                                          'publish', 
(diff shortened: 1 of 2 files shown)


src/events/mqtt.js
```diff
@@ -0,0 +1,109 @@
...
+    const options_internal = options;
+    const callback_internal = callback;
+
+    return {
+        topic_internal,
+        message_internal,
+        options_internal,
+        callback_internal
+    };
+};
```
galbash: no need for this if there is no messing with the parameters :)
we are doing this in other libraries since sometimes the callback is passed at a different argument (cause options is skipped for example) and we need to locate it. The logic here seems simple (unless options can be skipped, then this is an issue we should address)

src/events/mqtt.js
```diff
@@ -0,0 +1,92 @@
...
+ * from mqtt module
+ * @returns {Function} The wrapped function
+ */
+function publishWrapper(originalPublishFunc) {
+    return function internalPublishWrapper(topic, message, options, callback) {
+        let patchedCallback = callback;
+
+        try {
+            const { slsEvent: mqttEvent, startTime } = eventInterface.initializeEvent('MQTT',
+                this.options.host,
```
galbash: maybe let's switch topic and this one, what do you think?

src/events/mqtt.js
```diff
@@ -0,0 +1,92 @@
...
+        } catch (error) {
+            tracer.addException(error);
+        }
+        return originalConstructorFunc.apply(this, [options]);
+    };
+}
+
+module.exports = {
+    /**
+   * Initializes the AWS IOT tracer.
```
galbash: this seems wrong

galbash (dismissed):

src/events/mqtt.js
```diff
@@ -0,0 +1,92 @@
...
+            shimmer.wrap(
+                mqttClient,
+                'publish',
+                func => publishWrapper(func)
+            );
+            return mqttClient;
+        } catch (error) {
+            tracer.addException(error);
+        }
+        return originalConstructorFunc.apply(this, [options]);
```
galbash: there is a risk of calling the original function twice here (also i think you forgot streamBuilder)
better to declare mqttClient outside the try, and move the return to finally

src/events/mqtt.js
```diff
@@ -0,0 +1,92 @@
...
+    };
+}
+
+module.exports = {
+    /**
+   * Initializes the AWS MQTT tracer.
```
galbash: no need for AWS

galbash (approved):

PR #245, opened 2020-03-17
fix(mqtt): fix eslint and support mqtt subscribe function

haimrait:


package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "pretest": "./scripts/build_peg.sh",
     "test": "./scripts/run_tests.sh",
-    "lint:js": "eslint --max-warnings=0 ./src/ ./examples/ ./test/unit_tests ./index.js -f table --ext .js --ext .jsx",
+    "lint:js": "eslint --max-warnings=0 ./src/ ./examples/ ./test/unit_tests ./index.js -f table --ext .js --ext .jsx --fix",
```
galbash: why? this will mess up the CI cause people can commit unformatted code and the linter will fix it and pass.
you can add a `lint:js:fix`

src/events/mqtt.js
```diff
@@ -30,18 +30,18 @@ function publishWrapper(originalPublishFunc) {
                 message,
             };
             const promise = new Promise((resolve) => {
-                patchedCallback = () => {
+                patchedCallback = (err) => {
```
galbash: this actually seems like a problem i didn't notice before. we should pass all the arguments to the original callback (or we will hurt his application) - you can get `(err, ...rest)` and call `callback(err, ...rest)`

src/events/mqtt.js
```diff
@@ -53,6 +53,54 @@ function publishWrapper(originalPublishFunc) {
...
+                protocol: this.options.protocol,
+                host: this.options.host,
+            };
+            const payload = {
+                clientId: this.options.clientId,
+                protocolId: this.options.protocolId,
+                protocolVersion: this.options.protocolVersion,
+            };
+            const promise = new Promise((resolve) => {
+                patchedCallback = (err, granted) => {
```
galbash: same here, make sure there are no additional arguments

src/events/mqtt.js
galbash: if it just takes 2 arguments you can name them as is

galbash (changes requested):
few small comments

galbash (approved):

PR #246, opened 2020-03-18
Mqtt add support to on message

haimrait:


src/events/mqtt.js
```diff
@@ -97,10 +116,57 @@ function subscribeWrapper(originalSubscribeFunc) {
...
 }
 
+/**
+ * Wraps the on command function with tracing
+ * @param {Function} originalOnFunc The wrapped function
+ * from mqtt module
+ * @returns {Function} The wrapped function
+ */
+function onWrapper(originalOnFunc) {
+    return function internalOnWrapper(funcName, callback) {
```
galbash: 'eventName' might be better

src/events/mqtt.js
```diff
@@ -97,10 +116,57 @@ function subscribeWrapper(originalSubscribeFunc) {
...
+                            mqttEvent,
+                            startTime,
+                            null,
+                            responseMetadata,
+                            payload
+                        );
+                        if (callback) {
+                            callback(topic, message, ...rest);
+                        }
+                        tracer.addEvent(mqttEvent, promise);
```
galbash: no need for a promise here (you are not waiting for anything) - just remove the promise creation, resolve, and dont pass a promise parameter

galbash (changes requested):
two small comments

src/events/mqtt.js
haimrait: null or undefined instead?

galbash: undefined is good (just don't pass a second arg)

galbash (dismissed):

src/events/mqtt.js
```diff
@@ -1,8 +1,68 @@
...
+ * @param {*} message - mqtt message property.
+ * @returns {Object} options with epsagon id.
+ */
+function setEpsagonIdToMessage(epsagonId, message) {
+    if (typeof message === 'object') {
+        return { ...message, epsagonId };
+    } if (typeof message === 'string') {
+        try {
+            const resultMsg =
+            { ...JSON.parse(message), epsagonId };
```
galbash: I think `JSON.parse(message)` can sometimes return undefined when it fails instead of throwing and error, maybe let's validate that (also it can be a json dumped string, i.e.: `""this string""` where we do not want to add the id) by making sure after parsing that we got an object

src/events/mqtt.js
```diff
@@ -97,10 +162,58 @@ function subscribeWrapper(originalSubscribeFunc) {
...
+                    host: this.options.host,
+                };
+                const payload = {
+                    clientId: this.options.clientId,
+                    protocolId: this.options.protocolId,
+                    protocolVersion: this.options.protocolVersion,
+                };
+                patchedCallback = (topic, message, ...rest) => {
+                    const { slsEvent: mqttEvent, startTime } = eventInterface.initializeEvent('mqtt', topic, 'onMessage', 'mqtt');
+                    payload.message = message ? message.toString() : message;
```
galbash, line 194: are you sure about the `toString()` here (forgot to talk to you about it)? won't we get `[Object object]`? or is it never an object and always string / bytes?

haimrait: It always bytes

galbash (approved):

PR #247, opened 2020-03-21
feat(kafkajs.js): add support for kafkajs

ranrib:


src/consts.js
```diff
@@ -26,3 +26,6 @@ module.exports.MAX_LABEL_SIZE = 50 * 1024;
 module.exports.MAX_HTTP_VALUE_SIZE = 10 * 1024;
 
 module.exports.MAX_TRACE_SIZE_BYTES = 64 * 1024;
+
+// Key name to inject epsagon correlation ID
+module.exports.EPSAGON_HEADER = 'epsagon-id';
```
galbash: just make sure it will work with the java inst as well

src/events/kafkajs.js
```diff
@@ -0,0 +1,102 @@
...
+
+        const epsagonId = uuid4();
+        // eslint-disable-next-line no-param-reassign
+        messages.messages = messages.messages.map((message) => {
+            if (!message.headers) {
+                // eslint-disable-next-line no-param-reassign
+                message.headers = {};
+            }
+            // eslint-disable-next-line no-param-reassign
+            message.headers[EPSAGON_HEADER] = epsagonId;
```
galbash, line 43: you are setting the same id for all of the messages, this doesn't seem right

src/events/kafkajs.js
```diff
@@ -0,0 +1,102 @@
...
+            return message;
+        });
+
+        response = producer.originalSend(messages);
+
+        const sendPromise = new Promise((resolve) => {
+            let result;
+            let originalHandlerAsyncError;
+            response.then((res) => {
```
galbash: you are doing promise parallelizing here (which is not appropriate for this case) either call `originalSend` here or override response

src/events/kafkajs.js
```diff
@@ -0,0 +1,102 @@
...
+
+
+/**
+ * acts as a middleware for `producer.send()`
+ * @param {object} messages the messages param to send
+ * @param {Kafka} producer producer object
+ * @returns {Promise} The response promise
+ */
+function kafkaMiddleware(messages, producer) {
+    let response = Promise.resolve();
```
galbash: this is not good enough, I am afraid. check out other places we are mocking apis that return a promise and see how to handle it, an empty response is not what the client wants in case of an internal error on our end

src/events/kafkajs.js
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Wraps the kafkajs producer creation
+ * @param {Function} wrappedFunction The kafkajs producer function
+ * @returns {Function} The wrapped function
+ */
+function kafkaWrapper(wrappedFunction) {
+    return function internalKafkaWrapper(options) {
+        const producer = wrappedFunction.apply(this, [options]);
+        const patchedSend = messages => kafkaMiddleware(messages, producer);
```
galbash, line 106: because this is an arrow function, there is some this going lost magic that might happen here. might be best to use bind

src/events/kafkajs.js
```diff
@@ -0,0 +1,102 @@
...
+        producer.originalSend = producer.send;
+        producer.send = patchedSend;
```
galbash: bind should be used here:
```
producer.send = patchSend.bind(<w/e this should be, probably current scope's this)
```
please have a deeper look at it, as I am not an expert on `this` manipulation

galbash (changes requested):

galbash:
Also, can you make sure it's compatible with our Java lib?

src/events/kafkajs.js
ranrib: You're right. it's on purpose at the moment because we don't support setting multiple IDs on the same event. Any other alternative might require nice amount of re-design, and might be overkill for this stage

src/events/kafkajs.js
ranrib: 👍

src/events/kafkajs.js
ranrib: 👍

src/events/kafkajs.js
ranrib: It's not needed since its not using `this`, but added anyway 👍

src/events/kafkajs.js
```diff
@@ -0,0 +1,115 @@
...
+            }
+            // eslint-disable-next-line no-param-reassign
+            message.headers[EPSAGON_HEADER] = epsagonId;
+            return message;
+        });
+    } catch (err) {
+        tracer.addException(err);
+    }
+
+    const response = producer.originalSend(messages);
```
ranrib: add try catch and throw sync err

src/events/kafkajs.js
```diff
@@ -0,0 +1,115 @@
...
+    if (kafkaEvent) {
```
ranrib: move below

src/consts.js
ranrib: Done 👍

src/events/kafkajs.js
ranrib: The original function is an arrow function anyway:
image

src/events/kafkajs.js
```diff
@@ -0,0 +1,123 @@
...
+        });
+    } catch (err) {
+        tracer.addException(err);
+    }
+
+    try {
+        response = producer.originalSend(messages);
+    } catch (err) {
+        if (kafkaEvent) {
+            eventInterface.setException(kafkaEvent, err);
```
galbash, line 54: you never add the event in this case

src/events/kafkajs.js
```diff
@@ -0,0 +1,123 @@
...
+        }
+        throw err;
+    }
+
+    response.then((res) => {
```
galbash: you forgot to override response

src/events/kafkajs.js
ranrib: 👍

galbash (approved):

PR #253, opened 2020-03-29
fix(http.js): fix fetch-h2 response collision

### LINKED ISSUE #249: http response handling changes of epsagon 1.56.1 still produce runtime errors
even with the recent fixes of #244, we're still seeing errors with epsagon enabled, that are otherwise not a problem:

```
RangeError: Resource length mismatch (possibly incomplete body)
    at throwLengthMismatch (./node_modules/fetch-h2/dist/lib/body.js:17:11)
    at StreamResponse.validateIntegrity (./node_modules/fetch-h2/dist/lib/body.js:167:13)
    at awaitBuffer.then.already_1.tap.buffer 
```

could be caused by epsagon, consuming the chunks of the response.

ranrib:
1. Removing unused Wreck instrumentation (it wasn't used regardless of this PR).
2. Adding instrumentation to flag fetch-h2 library.
3. Adding instrumentation to `clientRequest.on('response')` -> `response.on('data')` to capture data in a non interfering way.
4. The shimmer upgrade is exactly to a fix we need to support instrumentation of async funcs inside classes (this is the only change in the lib - https://github.com/othiym23/shimmer/commit/ec15ba2bf2cffc25622fca0d277ea68f0ae2d8ee)

fixes https://github.com/epsagon/epsagon-node/issues/249

### DIFF
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -275,7 +275,9 @@ function httpWrapper(wrappedFunction) {
                 };
             }
 
-            shimmer.wrap(clientRequest, 'on', requestOnWrapper);
+            if (options && options.epsagonSkipResponseData) {
+                shimmer.wrap(clientRequest, 'on', requestOnWrapper);
+            }
 
             /**
              * Wraps 'write' method in a request to pick up request body
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -13,7 +13,6 @@ const serverlessEvent = require('../proto/event_pb.js');
 const eventInterface = require('../event.js');
 const errorCode = require('../proto/error_code_pb.js');
 const config = require('../config.js');
-const moduleUtils = require('./module_utils.js');
 const { MAX_HTTP_VALUE_SIZE } = require('../consts.js');
 const { isBlacklistURL, isBlacklistHeader } = require('../helpers/events');
 const {
@@ -418,18 +417,6 @@ function httpGetWrapper(module) {
     };
 }
 
-
-/**
- * Flagging fetch-h2 http1 requests with a flag to omit our response.on('data') because of collision
- * @param {Function} wrappedFunc connect function
- * @return {Function} the wrapped function
- */
-function fetchH2Wrapper(wrappedFunc) {
-    return function internalFetchH2Wrapper(options) {
-        return wrappedFunc.apply(this, [{ ...options, epsagonSkipResponseData: true }]);
-    };
-}
-
 module.exports = {
     /**
      * Initializes the http tracer
@@ -440,12 +427,5 @@ module.exports = {
         shimmer.wrap(http, 'request', httpWrapper);
         shimmer.wrap(https, 'get', () => httpGetWrapper(https));
         shimmer.wrap(https, 'request', httpWrapper);
-
-        moduleUtils.patchModule(
-            'fetch-h2/dist/lib/context-http1',
-            'connect',
-            fetchH2Wrapper,
-            fetch => fetch.OriginPool.prototype
-        );
     },
 };
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -57,6 +57,7 @@ function httpWrapper(wrappedFunction) {
         let url = a;
         let options = b;
         let callback = c;
+        const chunks = [];
         // handling case of request(options, callback)
         if (!(['string', 'URL'].includes(typeof url)) && !callback) {
             callback = b;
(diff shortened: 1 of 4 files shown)


src/events/http.js
```diff
@@ -215,6 +216,69 @@ function httpWrapper(wrappedFunction) {
...
+                return function internalResponseOnWrapper(resEvent, resCallback) {
+                    if (resEvent !== 'data' || typeof resCallback !== 'function') {
+                        return wrappedResFunction.apply(this, [resEvent, resCallback]);
+                    }
+                    const resPatchedCallback = (chunk) => {
+                        if (chunk) {
+                            const totalSize = chunks.reduce(
+                                (total, item) => item.length + total,
+                                0
+                            );
```
galbash: don't make the calculation each time, keep a sum

src/events/http.js
```diff
@@ -215,6 +216,69 @@ function httpWrapper(wrappedFunction) {
                 this, buildParams(url, options, patchedCallback)
             );
 
+            /**
+             * Wraps 'on' method in a response to capture data event.
+             * @param {Function} wrappedResFunction The wrapped end function
+             * @returns {Function} The wrapped function
+             */
+            function responseOnWrapper(wrappedResFunction) { // eslint-disable-line no-inner-declarations,max-len
```
galbash: must it be defined inside `httpWrapper`? it is getting very messy. better get `chunks` as a parameter (by adding another wrapping level) - not a must just my personal opinion, let me know if you think otherwise

galbash: same with the next one

src/events/http.js
```diff
@@ -215,6 +216,69 @@ function httpWrapper(wrappedFunction) {
...
+            }
+
+            /**
+             * Wraps 'on' method in a request to capture response event.
+             * @param {Function} wrappedReqFunction The wrapped end function
+             * @returns {Function} The wrapped function
+             */
+            function requestOnWrapper(wrappedReqFunction) { // eslint-disable-line no-inner-declarations,max-len
+                // epsagonMarker is sent only on our call in this module
+                return function internalRequestOnWrapper(reqEvent, reqCallback, epsagonMarker) {
```
galbash: might be better to set epsagonMarker to something more unique then 'true'

src/events/http.js
galbash: these eslint disables share my opinion I guess

src/events/http.js
```diff
@@ -215,6 +216,69 @@ function httpWrapper(wrappedFunction) {
...
+                        epsagonMarker ||
+                        typeof reqCallback !== 'function'
+                    ) {
+                        return wrappedReqFunction.apply(this, [reqEvent, reqCallback]);
+                    }
+                    const reqPatchedCallback = (res) => {
+                        if (res.EPSAGON_PATCH) {
+                            return reqCallback(res);
+                        }
+                        res.EPSAGON_PATCH = true;
```
galbash: i'd check res is not undefined just to be on the safe side

src/events/http.js
```diff
@@ -307,21 +369,25 @@ function httpWrapper(wrappedFunction) {
...
-                        }
-                    });
+                    // Listening to data only if options.epsagonSkipResponseData!=true or no options
+                    if (!options || (options && !options.epsagonSkipResponseData)) {
+                        res.on('data', (chunk) => {
+                            if (chunk) {
+                                const totalSize = chunks.reduce(
+                                    (total, item) => item.length + total,
+                                    0
+                                );
```
galbash: same

src/events/http.js
```diff
@@ -307,21 +369,25 @@ function httpWrapper(wrappedFunction) {
...
-                            return;
-                        }
-                        const totalSize = chunks.reduce((total, item) => item.length + total, 0);
-                        if (totalSize + chunk.length <= MAX_HTTP_VALUE_SIZE) {
-                            chunks.push(typeof chunk === 'string' ? Buffer.from(chunk) : chunk);
-                        }
-                    });
+                    // Listening to data only if options.epsagonSkipResponseData!=true or no options
+                    if (!options || (options && !options.epsagonSkipResponseData)) {
+                        res.on('data', (chunk) => {
```
galbash: this looks like shared logic that can be extracted

galbash (changes requested):
some small comments

src/events/http.js

src/events/http.js

src/events/http.js

src/events/http.js

src/events/http.js

src/events/http.js

galbash (approved):

Issue #256, opened 2020-04-07
http response handling changes of epsagon 1.58.3 still produce runtime errors when webpacked


tripodsan:
with #249 fixed, the tests now work locally, but when bundling with webpack, they still fail:

```
RangeError: Resource length mismatch (possibly incomplete body)
    at throwLengthMismatch (/Users/tripod/codez/helix/epsagon-testing/action-with-fetch/dist/default/epsagon-testing@1.0.2-bundle.js:4726:11)
    at StreamResponse.validateIntegrity (/Users/tripod/codez/helix/epsagon-testing/action-with-fetch/dist/default/epsagon-testing@1.0.2-bundle.js:4876:13)
    at awaitBuffer.then.already_1.tap.buffer (/Users/tripod/codez/helix/epsagon-testing/action-with-fetch/dist/default/epsagon-testing@1.0.2-bundle.js:4790:52)
    at /Users/tripod/codez/helix/epsagon-testing/action-with-fetch/dist/default/epsagon-testing@1.0.2-bundle.js:1571:15
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

when running the test locally, I see that the arguments passed into the `internalHttpWrapper` are:

```
http arguments [Arguments] {
  '0':
   { createConnection: [Function: createConnection],
     agent: false,
     hostname: 'adobeioruntime.net',
     method: 'GET',
     path:
      '/api/v1/web/helix/helix-services/word2md@v1...',
     port: 443,
     protocol: 'https:',
     headers: { connection: 'keep-alive' },
     epsagonSkipResponseData: true } }
```

but when running in the bundle, i get:

```
http arguments [Arguments] {
  '0':
   { createConnection: [Function: createConnection],
     agent: false,
     hostname: 'adobeioruntime.net',
     method: 'GET',
     path:
      '/api/v1/web/helix/helix-services/word2md@v1...,
     port: 443,
     protocol: 'https:',
     headers: { connection: 'keep-alive' } } }
```

It seems that the shimmer can't patch the method when it's bundled.

tripodsan:
the error is reproducible here:
https://github.com/tripodsan/epsagon-testing/tree/master/action-with-fetch

tripodsan:
as a quick workaround...could we add a general config option to `DisableHttpResponseBodyCapture` ?

PR #257, opened 2020-04-07
fix(http.js): add config to disable http responses capture

ranrib:
temp workaround for https://github.com/epsagon/epsagon-node/issues/256

### DIFF
--- a/src/events/http.js
+++ b/src/events/http.js
@@ -275,7 +275,11 @@ function httpWrapper(wrappedFunction) {
                 this, buildParams(url, options, patchedCallback)
             );
 
-            if (options && options.epsagonSkipResponseData) {
+            if (
+                options &&
+                options.epsagonSkipResponseData &&
+                !config.getConfig().disableHttpResponseBodyCapture
+            ) {
                 shimmer.wrap(
                     clientRequest,
                     'on',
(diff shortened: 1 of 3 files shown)


src/events/http.js
```diff
@@ -374,8 +378,10 @@ function httpWrapper(wrappedFunction) {
 
                 clientRequest.on('response', (res) => {
                     // Listening to data only if options.epsagonSkipResponseData!=true or no options
-                    if (!options || (options && !options.epsagonSkipResponseData)) {
-                        res.on('data', chunk => addChunk(chunk, chunks));
+                    if (!config.getConfig().disableHttpResponseBodyCapture) {
+                        if (!options || (options && !options.epsagonSkipResponseData)) {
+                            res.on('data', chunk => addChunk(chunk, chunks));
+                        }
```
galbash: use && ?

src/events/http.js
```diff
@@ -279,7 +279,11 @@ function httpWrapper(wrappedFunction) {
                 this, buildParams(url, options, patchedCallback)
             );
 
-            if (options && options.epsagonSkipResponseData) {
+            if (
+                options &&
+                options.epsagonSkipResponseData &&
```
tripodsan, line 284: shouldn't that be:

`if (options && options.epsagonSkipResponseData || config.getConfig().disableHttpResponseBodyCapture)` ?

ranrib: We want to enable this tracing for fetch-h2 only if all is true:
1. `options` exists.
2. `options.epsagonSkipResponseData` is set to true
3. Nobody disabled the response capture: `!config.getConfig().disableHttpResponseBodyCapture`

This handling is specific to fetch-h2. Later in the module, there's an handling for the rest of the libs.

galbash (approved):

src/events/http.js
tripodsan: but this is exactly what isn't working. the shimmer doesn't detect fetch-h2, and doesn't set the `options.epsagonSkipResponseData`. so we can force it with `disableHttpResponseBodyCapture`..maybe I don't understand the flags....

tripodsan: anyways...it seems to work now :-)

PR #263, opened 2020-04-27
feat(bunyan support): add bunyan logging instrumentation

adavidai:


src/events/bunyan.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Wrap bunyan logs
+ * @param {Function} wrappedFunction The function to wrap from bunyan
+ * @returns {function} emit wrapper function
+ */
+function emitWrapper(wrappedFunction) {
+    return function internalEmitWrapper(rec, ...args) {
+
+        if (tracer.getLogId() === null) {
```
galbash: might be better to call `getLogId()` once and store it in a variable

galbash: also might be better just checking `if(log_id)` and not comparing only to null

galbash (changes requested):

ronyis:
I think we should patch the logs only when the user configures to do so.
This flag should be also sent as part of the trace when set.

ranrib:
> I think we should patch the logs only when the user configures to do so.
> This flag should be also sent as part of the trace when set.

I would prefer to have a flag to disable it, and enable it by default for everyone

galbash (approved):

PR #275, opened 2020-05-05
feat(k8s): adding k8s container id

galbash:


src/containers/k8s.js
```diff
@@ -29,5 +42,9 @@ module.exports.loadK8sMetadata = function loadK8sMetadata() {
  */
 module.exports.addK8sMetadata = function addK8sMetadata(runner) {
     if (!runner || !k8sHostname) return;
-    eventIterface.addToMetadata(runner, { k8s_pod_name: k8sHostname });
+    eventIterface.addToMetadata(runner, {
+        is_k8s: true,
+        k8s_pod_name: k8sHostname,
+        k8s_container_id: k8sContainerId,
```
adavidai: Maybe only add it to metadata if it’s defined?

src/containers/k8s.js
```diff
@@ -18,6 +21,16 @@ module.exports.loadK8sMetadata = function loadK8sMetadata() {
...
         k8sHostname = os.hostname();
     }
+
+    if (!k8sContainerId) {
+        try {
+            const data = fs.readFileSync('/proc/self/cgroup');
+            const firstLineParts = data.toString('utf-8').split('\n')[0].split('/');
+            k8sContainerId = firstLineParts[firstLineParts.length - 1];
+        } catch (err) {
+            utils.debugLog('Error reading cgroup file', err);
```
adavidai: Maybe a more indicative error log of what won’t work - “Error loading k8s container id- cannot read cgroup file.”
(Maybe even add “ Show logs from span search will not work as expected”?)

galbash: i'll add the first line, but since this is only with `EPSAGON_DEBUG` on i don't think we need the second one

adavidai (dismissed):

adavidai (approved):

PR #276, opened 2020-05-05
fix(tracer.js): catching json stringify errors

### LINKED ISSUE #271: no traces are sent when JSON.stringify throws
When sending a trace in `sendCurrentTrace(traceSender)`,  a `traceJson` is assembled and later its size checked by `JSON.stringify`ing it:

```
    if (JSON.stringify(traceJson).length >= consts.MAX_TRACE_SIZE_BYTES) {
        // ...
    }
```

JSON.stringify can throw when the JSON has issues, such as a circular reference:

```
TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at Object.<anonymous> (<anonymous>:1:27)
    at sendCurrentTrace (/code/dist/getAllChannels/index.js:37495:5)
    at Promise.all.then (/code/dist/getAllChannels/index.js:37653:16)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Probably this error should be caught *and logged* :D and other traces allowed to be sent.

idonava:
Further to https://github.com/epsagon/epsagon-node/issues/271

### DIFF
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -320,7 +320,7 @@ function sendCurrentTrace(traceSender) {
     try {
         stringifyTraceJson = JSON.stringify(traceJson);
     } catch (err) {
-        utils.debugLog('Skip send traces.', err);
+        utils.debugWarn('Skip send traces.', err);
         return Promise.resolve();
     }
     if (stringifyTraceJson.length >= consts.MAX_TRACE_SIZE_BYTES) {
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -316,8 +316,14 @@ function sendCurrentTrace(traceSender) {
         platform: tracerObj.trace.getPlatform(),
     };
 
-
-    if (JSON.stringify(traceJson).length >= consts.MAX_TRACE_SIZE_BYTES) {
+    let stringifyTraceJson;
+    try {
+        stringifyTraceJson = JSON.stringify(traceJson);
+    } catch (err) {
+        utils.debugLog('Skip send traces.', err);
+        return Promise.resolve();
+    }
+    if (stringifyTraceJson.length >= consts.MAX_TRACE_SIZE_BYTES) {
         for (let attempt = 0; attempt <= MAX_STEPS; attempt += 1) {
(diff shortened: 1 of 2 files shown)


src/tracer.js
```diff
@@ -316,8 +316,14 @@ function sendCurrentTrace(traceSender) {
         platform: tracerObj.trace.getPlatform(),
     };
 
-
-    if (JSON.stringify(traceJson).length >= consts.MAX_TRACE_SIZE_BYTES) {
+    let stringifyTraceJson;
+    try {
+        stringifyTraceJson = JSON.stringify(traceJson);
```
solaris007, line 321: it looks like json.stringify is used in more locations than this. You could create a util or wrap system JSON. also, is a failure that prevents a trace from being sent worth a WARN level?

idonava: Thanks for your comment! I thought about it as well, but try/catch is a too expensive operation and we prefer not to do that every time.
About the JSON.stringify calls you mentioned - We just deleting properties from traceJson object (has already passed try/catch).
About the log levels (warn/error) - We will work on it in another PR soon.
Also, we are going to update this trimming system soon :)

src/tracer.js
```diff
@@ -316,8 +316,14 @@ function sendCurrentTrace(traceSender) {
...
+    } catch (err) {
+        utils.debugWarn('Skip send traces.', err);
```
galbash: can you change this to:
```
Epsagon - no trace was sent since there was an error serializing the trace. Please contact support.
```

src/utils.js
```diff
@@ -51,7 +51,7 @@ function reflectPromise(promise) {
 }
 
 /**
- * Prints a log if debugging is enabled
+ * Printing log if debugging is enabled
```
galbash: why?

src/tracer.js
galbash: I meant to actually use `console.warn` here (not just if `EPSAGON_DEBUG` is true) - if this happens the user should be alerted so he can contact us

galbash (approved):

PR #278, opened 2020-05-11
fix(lambda_env.js): disable_epsagon env on auto-tracing

idonava:


src/wrappers/lambda_env.js
```diff
@@ -59,5 +60,5 @@ function getUserHandler() {
         );
     }
 }
-
-module.exports.wrapper = lambda.lambdaWrapper(getUserHandler());
+module.exports.wrapper = config.getConfig().isEpsagonDisabled ? getUserHandler() :
```
galbash: the trinary operator will cause both expressions to be evaluated. better call `getUserHandler()` only once

idonava: fixed

galbash (approved):

PR #283, opened 2020-05-14
feat(openwhisk.js): capture namespace and host in openwhisk

ranrib:


src/events/openwhisk.js
```diff
@@ -69,7 +72,6 @@ function openWhiskWrapper(wrappedFunction) {
                     invokeEvent,
                     brief,
                     {
-                        activation_id: res.activationId,
```
tripodsan, line 72: why do you remove the response activation id?

ranrib: For some reason it was duplicated - it is already sent in the "brief"

galbash (approved):

PR #292, opened 2020-05-27
Adding winston logger + cw transport support

galbash:


src/events/winston.js
```diff
@@ -0,0 +1,56 @@
...
+    return function internalWriteWrapper(chunk, encoding, callback) {
+        if (!tracer.isLoggingTracingEnabled()) {
+            return wrappedFunction.apply(this, [chunk, encoding, callback]);
+        }
+        const traceId = tracer.getTraceId();
+        if (!traceId) {
+            return wrappedFunction.apply(this, [chunk, encoding, callback]);
+        }
+
+        if (!chunk || typeof chunk !== 'object') {
```
haimrait: All if statements can be in separate functions 
(you can merge them to one if statements, but it does not necessarily cause maybe it will be less readable so it up to you).

src/events/winston.js
```diff
@@ -0,0 +1,56 @@
...
+
+        const newChunk = {
+            epsagon: {
+                trace_id: traceId,
+            },
+        };
+        /* eslint-disable guard-for-in, no-restricted-syntax */
+        for (const key in chunk) {
+            newChunk[key] = chunk[key];
+        }
```
haimrait, line 37: for this case, you can do
const newChunk = {
           epsagon: {
                trace_id: traceId,
            },
            ... chunk
}

haimrait: or maybe because we have some node 6 issues,
maybe you can do it with Object.assign() approach.

src/events/winston.js
```diff
@@ -0,0 +1,56 @@
...
+
+        /* eslint-disable no-restricted-syntax */
+        for (const symbol of Object.getOwnPropertySymbols(chunk)) {
+            newChunk[symbol] = chunk[symbol];
+        }
```
haimrait, line 42: same as above

src/events/winston.js
galbash: can't because I want to copy non-enumerable symbols as well

src/events/winston.js
galbash: can't (Object.assign skip these)

haimrait (approved):

PR #304, opened 2020-06-05
fix(mongodb): improving mongodb connectionId parsing + verbose logging

galbash:


src/events/mongodb.js
```diff
@@ -24,21 +24,49 @@ function stringifyData(obj) {
...
 function onStartHook(event) {
     try {
         const startTime = Date.now();
+        utils.debugLog('Epsagon - MongoDB - handling event', event);
 
-        let { host, port } = event.connectionId;
-        if (!host) {
-            [host, port] = event.connectionId.split(':');
-        }
+        const { host, port } = getConnectionDetails(event ? event.connectionId : null);
```
ranrib, line 66: If there's no event is there any point to continue at all? Maybe check if event.connectionId?

src/events/mongodb.js
```diff
@@ -24,21 +24,49 @@ function stringifyData(obj) {
...
+ */
+function getConnectionDetails(connId) {
+    let host;
+    let port;
+    utils.debugLog('Epsagon - MongoDB - inspecting connectionId', connId);
+    if (connId) {
+        if (typeof connId === 'string') {
+            const parts = connId.split(':');
+            if (parts.length && parts[0][0] === '/') {
+                host = 'localhost';
```
ranrib, line 40: Can you add a quick explaination what is this scenario? (What's the original strong looks like)

src/events/mongodb.js
galbash: not really possible, just a safeguard

src/events/mongodb.js
galbash: in case it is a unix domain socket (i.e a file on the FS)

ranrib (approved):

tvaintrob (approved):

PR #306, opened 2020-06-09
feat(winston-cw): adding region and account id to winston cloudwatch

galbash:


src/events/winston_cloudwatch.js
```diff
@@ -1,9 +1,43 @@
 const Hook = require('require-in-the-middle');
 const utils = require('../utils');
+const tracer = require('../tracer.js');
```
tvaintrob: remove js

tvaintrob (approved):

PR #318, opened 2020-06-30
[EP-3515]: Fixing duplicate JSONify

henperez:


src/wrappers/node.js
```diff
@@ -39,6 +39,7 @@ function createRunner(functionToWrap, args) {
     runner.setResource(runnerResource);
     eventInterface.addToMetadata(runner, {
         args_length: args.length,
+        args: args,
```
ranrib: That should be in the payload section

galbash (dismissed):

galbash (approved):

PR #320, opened 2020-07-06
fix(http.js): update got module handling

ranrib:


src/events/http.js
```diff
@@ -126,6 +126,14 @@ function httpWrapper(wrappedFunction) {
             options = null;
         }
 
+        // handling case of got.post(url, options)
+        if (a.constructor && a.constructor.name === 'URL' && typeof b === 'object' && !c) {
```
galbash: I think it's time to create a `parseArguments` function

galbash (changes requested):

galbash (approved):

PR #332, opened 2020-07-21
feat(lables): support labels inside init

haimrait:


src/tracer.js
```diff
@@ -524,17 +533,7 @@ module.exports.sendTraceSync = function sendTraceSync() {
...
-        utils.debugLog('Failed to label without an active tracer');
-        return;
-    }
-    let labels = {};
-    labels[key] = value;
-    labels = utils.flatten(labels);
-    Object.keys(labels).forEach((k) => {
-        eventInterface.addLabelToMetadata(tracerObj.currRunner, k, labels[k]);
-    });
+    config.setConfig({...config.getConfig(), labels: [...config.getConfig().labels, {[key]: value}]})
```
ranrib: You've removed the ability to provide complex objects into add label - please return it.

src/tracer.js
```diff
@@ -241,6 +241,13 @@ function getTrimmedTrace(traceSize, jsTrace) {
     return trimmedTrace;
 }
 
+function setLabelsToEvent(tracerObj) {
+    config.getConfig().labels.forEach((label) => {
```
ranrib: you can use `entries` to get both key and value at the same time

src/tracer.js
```diff
@@ -241,6 +241,13 @@ function getTrimmedTrace(traceSize, jsTrace) {
     return trimmedTrace;
 }
 
+function setLabelsToEvent(tracerObj) {
```
ranrib: I would use a more correct name like `addLabelsToTrace` or `setLabelsToRunner`

src/tracer.js
```diff
@@ -251,6 +258,7 @@ function getTrimmedTrace(traceSize, jsTrace) {
  */
 function sendCurrentTrace(traceSender) {
     const tracerObj = module.exports.getTrace();
+    setLabelsToEvent(tracerObj)
```
ranrib: I would move this and the next one a few lines below, since tracerObj can be `undefined`

README.md
```diff
@@ -95,6 +95,15 @@ epsagon.label('key', 'metric')
 epsagon.label('itemsInCart', itemsInCart)
 ```
 
+You can also use it iniside init function:
+```javascript
+epsagon.init({
+  token: 'epsagon-token',
+  appName: 'app-name-stage',
+  labels: [{'key', 'value'}, {'userId', userId}],
```
galbash: Should probably get an object / array of arrays (couples)

src/config.js
```diff
@@ -166,4 +167,8 @@ module.exports.setConfig = function setConfig(configData) {
     if (Number(configData.sendTimeout)) { // we do not allow 0 as a timeout
         config.sendTimeout = Number(configData.sendTimeout);
     }
+
+    if (configData.labels) {
+        config.labels = [...configData.labels];
```
galbash: is it an array or an object? (the default is initialized as an object, but here you set it as an array)

src/tracer.js
```diff
@@ -524,16 +537,11 @@ module.exports.sendTraceSync = function sendTraceSync() {
...
-    if (!tracerObj || !tracerObj.currRunner) {
-        utils.debugLog('Failed to label without an active tracer');
-        return;
-    }
-    let labels = {};
-    labels[key] = value;
-    labels = utils.flatten(labels);
-    Object.keys(labels).forEach((k) => {
-        eventInterface.addLabelToMetadata(tracerObj.currRunner, k, labels[k]);
+    config.setConfig({
```
galbash: This change is not valid (we want to preserve the behavior that `addLabel` is per trace basis)

src/tracer.js
```diff
@@ -472,6 +484,7 @@ module.exports.postTrace = function postTrace(traceObject) {
  */
 module.exports.sendTrace = function sendTrace(runnerUpdateFunc) {
     const tracerObj = module.exports.getTrace();
+    setLabelsToEvent(tracerObj);
```
galbash: you can move this to where we set the ecs metadata instead of doing it twice (here and `sendCurrentTrace`)

galbash (changes requested):
few comments

galbash (approved):

PR #337, opened 2020-07-30
fix(utils.js): disable events in lambda env

ranrib:


src/config.js
```diff
@@ -43,7 +43,7 @@ const config = {
     sendTimeout: (Number(process.env.EPSAGON_SEND_TIMEOUT_SEC) || DEFAULT_TIMEOUT_SEC) * 1000.0,
     decodeHTTP: (process.env.EPSAGON_DECODE_HTTP || 'TRUE').toUpperCase() === 'TRUE',
     disableHttpResponseBodyCapture: (process.env.EPSAGON_DISABLE_HTTP_RESPONSE || '').toUpperCase() === 'TRUE',
-    loggingTracingEnabled: (process.env.EPSAGON_LOGGING_TRACING_ENABLED || 'TRUE').toUpperCase() === 'TRUE',
+    loggingTracingEnabled: !utils.isLambdaEnv && (process.env.EPSAGON_LOGGING_TRACING_ENABLED || 'TRUE').toUpperCase() === 'TRUE',
```
galbash: please move this check to when selecting the default value and not overriding the final result (if a user explicitly set `EPSAGON_LOGGING_TRACING_ENABLED` it should be set to true)

galbash (changes requested):

henperez (approved):

galbash (approved):

PR #338, opened 2020-07-30
fix(tracer): fix sampling on lambda

liadgo:


src/tracer.js
```diff
@@ -49,6 +49,7 @@ module.exports.createTracer = function createTracer() {
         trace: tracerObj,
         currRunner: null,
         pendingEvents: new Map(),
+        sampleRate: config.getConfig().sampleRate,
```
galbash: why set this on a per trace basis?

tvaintrob (dismissed):

test/unit_tests/wrappers/test_openwhisk_traces.js
```diff
@@ -46,9 +46,9 @@ describe('openwhiskWrapper tracer tests', () => {
             'addException'
         );
 
-        this.postTraceStub = sinon.stub(
+        this.sendTraceStub = sinon.stub(
```
galbash: why is this change required? did tests failed without it?

src/tracer.js
```diff
@@ -364,6 +370,10 @@ function sendCurrentTrace(traceSender) {
 
     const sendResult = traceSender(traceJson);
     tracerObj.pendingEvents.clear();
+
+    if (tracerObj.sampleRate !== consts.DEFAULT_SAMPLE_RATE) {
+        tracerObj.deleted = true;
+    }
```
galbash: you can just check the config here instead of doing it on a per-trace basis (still set deleted on the current trace)

test/unit_tests/wrappers/test_openwhisk_traces.js
liadgo: Yeah.
actually the entire test seems to be buggy.
TraceObj is not being initialised, therefore - no runner is being set (look at addRunner func), and the trace is will not be posted from now on  (which is a GOOD behavior) 
will fix that in the future.  @ranrib please approve.

src/tracer.js
liadgo: Had this thought. for somereason chose to do this per-trace basis.
Looks ok with config. :)
changing!

galbash (approved):

PR #343, opened 2020-08-10
fix(config.js): disable tracing on config set

ranrib:


src/config.js
```diff
@@ -109,6 +109,10 @@ module.exports.setConfig = function setConfig(configData) {
         config.token = configData.token;
     }
 
+    if (configData.isEpsagonDisabled) {
```
galbash, line 112: why not just `disable`?

galbash (changes requested):
naming question

src/config.js
ranrib: `isEpsagonDisabled` is already defined but not working

galbash (approved):

PR #354, opened 2020-09-09
feat(kafka-node.js): support kafka-node

ranrib:


src/events/kafka-node.js
```diff
@@ -0,0 +1,126 @@
...
+        } catch (err) {
+            tracer.addException(err);
+        }
+
+        const responsePromise = new Promise((resolve) => {
+            patchedCallback = (err, data) => {
+                let callbackResult;
+                try {
+                    if (callback && typeof callback === 'function') {
+                        callbackResult = callback(err, data);
```
galbash: better call finalize event before callback (for time measurement)

galbash: maybe just call the callback in a `finally` clause

galbash (changes requested):
few small comments

src/events/kafka-node.js
ranrib: 👍

src/events/kafka-node.js
```diff
@@ -0,0 +1,127 @@
...
+                    );
+                } catch (callbackErr) {
+                    tracer.addException(callbackErr);
+                } finally {
+                    if (callback && typeof callback === 'function') {
+                        callbackResult = callback(err, data);
+                    }
+                }
+                resolve();
+                return callbackResult;
```
galbash, line 94: you can move the `resolve()` to the finally and then you won't need `callbackResult` at all

galbash (approved):

PR #362, opened 2020-09-23
fix(lambda.js): support async step function response

ranrib:


src/wrappers/lambda.js
```diff
@@ -316,41 +316,49 @@ module.exports.lambdaWrapper = function lambdaWrapper(functionToWrap) {
...
+
+/**
+ * Extract the step data from the event, and updates the runner and the response.
+ * @param {object} originalEvent Lambda function event data
+ * @param {Event} runner event
+ * @param {object} response Lambda function response data
+ * @returns {object} response the update response with steps
+ */
+function extractAndAppendStepData(originalEvent, runner, response) {
+    let step = null;
```
haimrait, line 328: what will happen if response is err?

ranrib: We're only getting to this point if no error returned / raised

haimrait (approved):

PR #365, opened 2020-09-26
fix(http.js): don't capture standard ports

ranrib:


README.md
```diff
@@ -512,6 +512,14 @@ const messageHandler = message => {
 };
 ```
 
+Or in batch message handler:
+```javascript
+const batchMessageHandler = messages => {
+  messages[0].epsagon.label('key', 'value');
```
haimrait: messages[0] ?

haimrait (approved):

PR #366, opened 2020-09-28
fix(module_utils.js): add missing node_modules in Lambda envs

ranrib:


src/consts.js
```diff
@@ -34,6 +34,9 @@ module.exports.DEFAULT_SAMPLE_RATE = 1;
 // Key name to inject epsagon correlation ID
 module.exports.EPSAGON_HEADER = 'epsagon-trace-id';
 
+// In some cases we manually add the Lambda node_modules path, where it is not found by default
+module.exports.LAMBDA_DEFAULT_NODE_MODULES_PATH = '/var/task/node_modules';
```
haimrait, line 38: the user can change it if he wants? with the environment variable?

src/events/module_utils.js
```diff
@@ -53,15 +55,20 @@ module.exports.getModules = function getModules(id) {
...
             autoNodePaths = getAllNodeModulesPaths(rootFolder);
         }
+        utils.debugLog('Found the following paths', autoNodePaths);
         autoNodePaths.forEach((nodePath) => {
             if (!searchPaths.includes(nodePath)) {
                 searchPaths.push(nodePath);
             }
         });
     }
+    if (utils.isLambdaEnv && !searchPaths.includes(LAMBDA_DEFAULT_NODE_MODULES_PATH)) {
```
haimrait, line 65: this is the case if the user provides an external path to node_modules?

haimrait (approved):

src/events/module_utils.js
ranrib: This is for all cases, where we are running inside Lambda - to make sure we're looking at /var/task/node-modules

PR #367, opened 2020-09-28
fix(lambda.js): fix response alter in case of null or array

ranrib:


src/wrappers/lambda.js
```diff
@@ -28,7 +28,12 @@ module.exports.FAILED_TO_SERIALIZE_MESSAGE = FAILED_TO_SERIALIZE_MESSAGE;
  * @returns {object} returnValue Updated return value
  */
 function propagateEpsagonId(runner, returnValue) {
-    if (process.env.EPSAGON_PROPAGATE_LAMBDA_ID && typeof returnValue === 'object') {
+    if (
+        process.env.EPSAGON_PROPAGATE_LAMBDA_ID &&
+        returnValue &&
+        !Array.isArray(returnValue) &&
```
haimrait, line 34: if the return value is an array
you can set and prop on them
arr.prop = 'value';

ranrib: But this is for a Lambda response - so it's for an actual JSON that we return, and not just internal to the code

haimrait (approved):

PR #373, opened 2020-10-08
feat(batch processing): traces queue + batch sending + byte size limit

itaykat:


src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
+/* eslint-disable prefer-destructuring */
+/* eslint-disable no-console */
```
galbash: this sounds wrong

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+/* eslint-disable valid-jsdoc */
+/* eslint-disable require-jsdoc */
```
galbash: why not just add the docs?

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
+/* eslint-disable prefer-destructuring */
```
galbash: is there any reason to not prefer destructing?

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+   * Init queue event listners
+   */
+    initQueue() {
+        this.removeAllListeners();
+        this.currentByteSize = 0;
+        this.flush();
+        //  check if queue reached batch size
+        this.on('traceQueued', () => {
```
galbash: I think that using event emitter makes it so you can send a couple of batches in parallel - not sure this is what you want

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+            if (this.batchSizeReached()) {
+                utils.debugLog(`Queue size reached ${this.currentSize}, batch ready! Releasing... `);
+                this.releaseBatch();
+            }
+        });
+
+        // update queue size in bytes + check if limit reached
+        this.on('traceQueued', (trace) => {
+            this.addToCurrentByteSize(trace);
```
galbash: calling `JSON.stringify` of a trace is expensive, better if we can get another estimate. If not might be better to do it one and then generating the final list with `'[' + traces.join(',') + ']'`.

galbash: also better to check here curr size + next instead of adding before the check, to prevent having releaseBatch get a parameter which increases complaxity

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+        return this.traces.length;
+    }
+
+
+    /**
+   * Checks if queue size reached batch size
+   * @returns {Boolean}
+   */
+    batchSizeReached() {
+        return this.currentSize === this.batchSize;
```
galbash: don't you mean >=?

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+        this.emit('traceQueued', trace);
+        return this;
+    }
+
+
+    /**
+   * Release batch of traces
+   * @returns {TraceQueue}
+   */
+    releaseBatch(count = this.batchSize) {
```
galbash: what's the added value of breaking `releaseBatch` & `batchReleased` to 2 different async actions? sounds like there is no reason to do it if we don't want concurrent sends.

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+});
+
+/**
+ * Post given batch to epsagon's infrastructure.
+ * @param {*} batchObject The batch data to send.
+ * @returns {Promise} a promise that is resolved after the batch is posted.
+ */
+function postBatch(batchObject) {
+    utils.debugLog(`Posting batch to ${config.getConfig().traceCollectorURL}...`);
+    utils.debugLog(`Batch: ${JSON.stringify(batchObject, null, 2)}`);
```
galbash: this is a very expensive debug log, maybe print it only if debug is on (even if it is not printed the call to JSON.stringify is executed)

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+    httpAgent: new http.Agent({ keepAlive: true }),
+    httpsAgent: new https.Agent({ keepAlive: true }),
+});
+
+/**
+ * Post given batch to epsagon's infrastructure.
+ * @param {*} batchObject The batch data to send.
+ * @returns {Promise} a promise that is resolved after the batch is posted.
+ */
+function postBatch(batchObject) {
```
galbash: how is this different than the regular postTrace? maybe we can just export it to a different module `https_sender` and use the same function?

src/trace_queue.js
```diff
@@ -0,0 +1,188 @@
...
+}
+
+
+/**
+ * Session for the post requests to the collector
+ */
+const session = axios.create({
+    timeout: config.getConfig().sendTimeout,
+    httpAgent: new http.Agent({ keepAlive: true }),
+    httpsAgent: new https.Agent({ keepAlive: true }),
```
haimrait: i think better to add token authorization header in the session object for all requests.
you can do it in the init function

src/trace_queue.js
itaykat: so basically the postBatch() is just a duplicated and customised postTrace() (which is a function already exists in the tracer), I'll definitely remove unnecessary debug logs once we'll get stable. Also took your advice to log only on DEBUG mode.

src/trace_queue.js
itaykat: I've duplicated for the ease of development, once we'll get stable we can think of joining them.

src/trace_queue.js
ronyis: Yep, it doesn't seem like a good solution for this case.
We need that on reaching batch size / time the buffered traces will get sent asynchronously. Any other traces that are buffered should not be sent until sending the previous batch ends.

examples/batch_example.js
```diff
@@ -0,0 +1,58 @@
+const epsagon = require('../src/index');
+const http = require('http');
+
+
+epsagon.init({
+    token: process.env.EPSAGON_TOKEN,
+    appName: 'batch-test',
+    metadataOnly: false,
+    sendBatch: true,
+    batchSize: 1,
```
ronyis: Looks weird to use that in this example

examples/batch_example.js
```diff
@@ -0,0 +1,58 @@
...
+    maxBatchSizeBytes: 5000000
```
ronyis: IMO better to use in KB (and to update the var name in this case)

examples/batch_example.js
```diff
@@ -0,0 +1,58 @@
...
+function testSyncFunction() {
+    console.log("logging something")
+}
+
+const wrappedAsyncTestFunction = epsagon.nodeWrapper(testAsyncFunction);
+const wrappedSyncTestFunction = epsagon.nodeWrapper(testSyncFunction);
+
+async function main (){
+    await wrappedAsyncTestFunction()
+    // await wrappedAsyncTestFunction()
```
ronyis: Remove?

src/config.js
```diff
@@ -44,6 +44,12 @@ const config = {
     decodeHTTP: (process.env.EPSAGON_DECODE_HTTP || 'TRUE').toUpperCase() === 'TRUE',
     disableHttpResponseBodyCapture: (process.env.EPSAGON_DISABLE_HTTP_RESPONSE || '').toUpperCase() === 'TRUE',
     loggingTracingEnabled: (process.env.EPSAGON_LOGGING_TRACING_ENABLED || (!utils.isLambdaEnv).toString()).toUpperCase() === 'TRUE',
+    sendBatch: (process.env.EPSAGON_SEND_BATCH || 'FALSE').toUpperCase() === 'TRUE',
+    batchSize: (Number(process.env.EPSAGON_BATCH_SIZE) || consts.DEFAULT_BATCH_SIZE),
```
ronyis, line 48: We should have an upper limit on batch sizes (that should probably match the ones that will be configured in TC)

src/trace_queue.js
itaykat: That is a good point, the issue here that we can have some kind of race condition... let's talk about that.

ronyis:
I think that some important things are missing:
1. Are you using `maxTraceWait`?
2. There isn't any limit on keeping to collect traces, which can probably result in a memory error.
3. As mentioned in other comments - there isn't a real queue for sending traces

src/trace_queue.js
```diff
@@ -0,0 +1,196 @@
...
+    byteSizeLimitReached() {
+        return this.currentByteSize >= this.maxBatchSizeBytes;
+    }
+
+    /**
+   * add given trace byte size to total byte size
+   * @returns {Number}
+   */
+    addToCurrentByteSize(trace) {
+        this.currentByteSize += JSON.stringify(trace).length;
```
ronyis: `JSON. stringify` can be CPU expensive. Maybe we should save the traces as JSONs already?

src/trace_queue.js
itaykat: That's relate to the releaseBatch comment below, let's talk about it

examples/batch_example.js
itaykat: We already use Bytes in the regular sending method... that's for the sake of uniformity

src/config.js
itaykat: So you say the only limit for batch size will be a constant? not user configured?

src/trace_queue.js
itaykat: Different PR.

src/config.js
itaykat: I'll define softLimit (user defined) + hardLimit (equal to TC limit)

src/trace_queue.js
itaykat: @ronyis so the releaseBatch called Synced, while the sendBatch is Asynced

src/config.js
ronyis: 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "epsagon",
-  "version": "0.0.0-development",
+  "version": "1.88.1-beta.2",
```
enoodle: Maybe this is a mistake?

haimrait (approved):

PR #379, opened 2020-10-18
feat(tracer): trimming exceptions when tracer is more than max size

idonava:


amitlicht (dismissed):

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
 function getTrimmedTrace(traceSize, jsTrace) {
     let currentTraceSize = traceSize;
     const trimmedTrace = Object.assign({}, jsTrace);
-    trimmedTrace.events = jsTrace.events.sort(event => (['runner', 'trigger'].includes(event.origin) ? -1 : 1));
+    let totalTrimmedExceptions = 0;
```
amitlicht, line 220: general note: try to keep it to 1 PR per 1 feature/fix (meaning this should've been a separate PR). Also don't forget to fix the PR title / commit message to take this change into account.

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
 function getTrimmedTrace(traceSize, jsTrace) {
```
amitlicht, line 217: this function is becoming too long (and too nested). Consider moving every "if" block content to its own dedicated function (`trimTraceExceptions`, `trimTraceEventsMetadata`, `trimTraceEvents`, etc)

src/tracer.js
```diff
@@ -236,9 +247,9 @@ function getTrimmedTrace(traceSize, jsTrace) {
                 }
             }
         }
-        if (totalTrimmedEvents) {
-            utils.debugLog(`Epsagon - Trace size is larger than maximum size, ${totalTrimmedEvents} events was trimmed.`);
-        }
+    }
+    if (totalTrimmedEvents || totalTrimmedExceptions) {
+        utils.debugLog(`Epsagon - Trace size is larger than maximum size, ${totalTrimmedEvents} events and ${totalTrimmedExceptions} exceptions was trimmed.`);
```
amitlicht: "... were trimmed."

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
...
-            const originalEventMetadataSize = JSON.stringify(eventMetadata).length;
-            const trimmedMetadata = getTrimmedMetadata(eventMetadata, isRunner);
-            if (trimmedMetadata) {
-                eventMetadata = trimmedMetadata;
-                const trimmedSize =
+    if (currentTraceSize >= consts.MAX_TRACE_SIZE_BYTES) {
+        trimmedTrace.events = jsTrace.events.sort(event => (['runner', 'trigger'].includes(event.origin) ? -1 : 1));
+        for (let i = jsTrace.events.length - 1; i >= 0; i -= 1) {
+            const currentEvent = trimmedTrace.events[i];
+            const isRunner = currentEvent.origin === 'runner';
```
amitlicht: `isRunner` is needed only in the `if (eventMetadata)` block, so move it inside the if block.

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
...
-        let eventMetadata = currentEvent.resource.metadata;
-        if (eventMetadata) {
-            const originalEventMetadataSize = JSON.stringify(eventMetadata).length;
-            const trimmedMetadata = getTrimmedMetadata(eventMetadata, isRunner);
-            if (trimmedMetadata) {
-                eventMetadata = trimmedMetadata;
-                const trimmedSize =
+    if (currentTraceSize >= consts.MAX_TRACE_SIZE_BYTES) {
+        trimmedTrace.events = jsTrace.events.sort(event => (['runner', 'trigger'].includes(event.origin) ? -1 : 1));
+        for (let i = jsTrace.events.length - 1; i >= 0; i -= 1) {
```
amitlicht, line 234: IMO this would be easier to read if you sorted the events in reverse order, and then the for loop is over increasing i.

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
...
+            const currentEvent = trimmedTrace.events[i];
+            const isRunner = currentEvent.origin === 'runner';
+            let eventMetadata = currentEvent.resource.metadata;
+            if (eventMetadata) {
+                const originalEventMetadataSize = JSON.stringify(eventMetadata).length;
```
amitlicht, line 239: this is used only in the `if (trimmedMetadata)` block, so move it in.

src/tracer.js
amitlicht: Consider adding logging for trimmed metadata as well.

src/tracer.js
```diff
@@ -202,29 +202,40 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
 function getTrimmedTrace(traceSize, jsTrace) {
     let currentTraceSize = traceSize;
     const trimmedTrace = Object.assign({}, jsTrace);
-    trimmedTrace.events = jsTrace.events.sort(event => (['runner', 'trigger'].includes(event.origin) ? -1 : 1));
+    let totalTrimmedExceptions = 0;
+    let totalTrimmedEvents = 0;
+    // Trimming trace exceptions.
+    if (trimmedTrace.exceptions.length > 1) {
```
amitlicht, line 223: there's no `if (currentTraceSize >= consts.MAX_TRACE_SIZE_BYTES)` here which I found a bit weird. 
does this need to happen even if currentTraceSize is small enough?

amitlicht (changes requested):

src/tracer.js
idonava: you're right, I'll export the new STRONG_ID_KEYS params commit to a new PR.

src/tracer.js
idonava: The first check exists before the calling to this function.

src/tracer.js
idonava: Actually, you're right, but it's for another PR.

src/tracer.js
idonava: moved

src/tracer.js
idonava: eventMetaData changes inside getTrimmedMetadata (we can fix it by deepcopy of eventMetaData, but for another PR).

src/tracer.js
idonava: There is an indication (is_trimmed) inside the events when we're trimming event metadata.

src/tracer.js
amitlicht: in that case it's a problematic choice of naming for `getTrimmedMetadata`  (not this PR's fault, but still)

src/tracer.js
```diff
@@ -202,29 +217,41 @@ function getTrimmedMetadata(eventMetadata, isRunner) {
...
+        const [firstException, totalTrimmed, reduceSize] = trimTraceExceptions(
+            trimmedTrace.exceptions
+        );
+        currentTraceSize -= reduceSize;
+        totalTrimmedExceptions = totalTrimmed;
+        trimmedTrace.exceptions = [firstException];
```
amitlicht, line 229: should happen inside `trimTraceExceptions` (and then avoid returning firstException just for that).

amitlicht (approved):

PR #380, opened 2020-10-18
feat(sqs): supporting multiple sqs messages for lambda trigger

enoodle:


src/triggers/aws_lambda.js
```diff
@@ -111,11 +111,11 @@ function createSQSTrigger(event, trigger) {
     resource.setOperation('ReceiveMessage');
     const sqsMessageBody = event.Records[0].body || '{}';
     eventInterface.addToMetadata(trigger, {
-        'MD5 Of Message Body': event.Records[0].md5OfBody,
-        Attributes: event.Records[0].attributes,
-        'Message Attributes': event.Records[0].messageAttributes,
+        'MD5 Of Message Body': event.Records.map(r => r.md5OfBody),
+        Attributes: event.Records.map(r => r.attributes),
+        'Message Attributes': event.Records.map(r => r.messageAttributes),
```
haimrait: why the keys have spaces?

enoodle: These are the original key names, I don't want to change them in case these strings are used somewhere else. They are similar to the ones in other instrumentaitons.

ronyis (dismissed):

haimrait (approved):

PR #385, opened 2020-11-08
feat(sql): collecting response data

henperez:


src/tracer.js
```diff
@@ -182,7 +182,7 @@ module.exports.restart = function restart() {
 function getTrimmedMetadata(eventMetadata, isRunner) {
     let trimmedEventMetadata;
     Object.keys(eventMetadata).forEach((eventKey) => {
-        if (!isStrongId(eventKey) && (isRunner && eventKey !== 'labels')) {
+        if (!(isStrongId(eventKey) || (isRunner && eventKey == 'labels'))) {
```
henperez: _Fixing a bug that prevented us from trimming events' metadata correctly_

src/events/sql.js
```diff
@@ -96,9 +96,13 @@ module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callbac
                 if (err) {
                     eventInterface.setException(dbapiEvent, err);
                 } else {
-                    let { rowCount } = res;
+                    let { rowCount, rows } = res;
```
amitlicht, line 101: This would be nicer (and clearer) if calling `addToMetadata(... { 'sql.rows': ...});` would happen in a single place. 
Something like that: 
```
let { rowCount, rows } = res;
if (!rowCount && res instanceof Array) {
    rowCount = res.length
    rows = res
}

eventInterface.addToMetadata(dbapiEvent, { rowCount });
if (rowCount && rows instanceof Array && rows.length) {
    eventInterface.addToMetadata(dbapiEvent, { 'sql.rows': rows });
}
```

amitlicht (changes requested):

src/events/sql.js
henperez: Done

amitlicht (dismissed):

src/events/sql.js
```diff
@@ -96,11 +96,15 @@ module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callbac
...
                     if (!rowCount && res instanceof Array) {
                         rowCount = res.length;
+                        rows = res;
+                    }
+                    eventInterface.addToMetadata(dbapiEvent, { rowCount });                 
+                    if (rowCount && rows instanceof Array && rows.length) {
+                        eventInterface.addToMetadata(dbapiEvent, { 'sql.rows': rows });
```
ronyis: It's not good that there is no trimming here:
- When the maximum value is reached other (small) tags could be removed
- The `rows` object in many cases could be **very** big, and the process of handling it (before the trimming) could be wasteful

src/tracer.js
```diff
@@ -182,7 +182,7 @@ module.exports.restart = function restart() {
 function getTrimmedMetadata(eventMetadata, isRunner) {
     let trimmedEventMetadata;
     Object.keys(eventMetadata).forEach((eventKey) => {
-        if (!isStrongId(eventKey) && (isRunner && eventKey !== 'labels')) {
+        if (!(isStrongId(eventKey) || (isRunner && eventKey === 'labels'))) {
```
ronyis, line 185: 🙏

src/events/sql.js
henperez: Trying to trim it is very expensive if not needed. I tried this with 100 rows and it's fine, it's all data dependant. Either way the code is trimming before sending the trace

ronyis: I think that the code could be expensive before the trimming takes place

henperez: Why? we just point the reference to an existing object. It's the same if we'll stringify 1M lines and trim it one-by-one (which will result in more stringify operations of less data) and then stringify the trace without 1M lines than stringify 1M+trace at the end.

henprtz:
@ronyis - Added limitation for 100 rows

henperez:
@ronyis & @amitlicht - I've added limitation for 100 rows to reduce performance hit

haimrait (approved):

PR #389, opened 2020-11-16
feat(ec2): capture ec2 metadata

idonava:


src/containers/ec2.js
```diff
@@ -0,0 +1,44 @@
...
+module.exports.loadEC2Metadata = function loadEC2Metadata() {
+    if (currentEC2Labels) return Promise.resolve(currentEC2Labels);
+    const promises = [];
+    utils.debugLog('Loading EC2 metadata');
+    attributeToGet.forEach((attribute) => {
+        promises.push(axios.get(URL + attribute).then((response) => {
+            utils.debugLog(`Received response for ${attribute}: ${response.data}`);
+            if (response.status === 200) {
+                if (!currentEC2Labels) currentEC2Labels = {};
+                currentEC2Labels[`ec2.${attribute}`] = response.data;
```
ranrib: I think we should do `aws.ec2.attribute`

src/containers/ec2.js
```diff
@@ -0,0 +1,44 @@
+const axios = require('axios');
+const utils = require('../utils');
+const eventIterface = require('../event');
+
+let currentEC2Labels = null;
+
+const URL = 'http://169.254.169.254/latest/meta-data/';
```
haimrait, line 7: why is it a static ip?

src/containers/ec2.js
idonava: You're right.

src/containers/ec2.js
idonava: it's always the same IP to get the metadata in the running instance

src/containers/ec2.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Load EC2 metadata and store it
+ * @returns {Promise} when resolved will contain the EC2 metadata
+ */
+module.exports.loadEC2Metadata = function loadEC2Metadata() {
+    if (currentEC2Labels) return Promise.resolve(currentEC2Labels);
+    const promises = [];
+    utils.debugLog('Loading EC2 metadata');
+    attributeToGet.forEach((attribute) => {
+        promises.push(axios.get(URL + attribute).then((response) => {
```
haimrait: add a timeout to axios request, otherwise, it can take a lot of time until rejection

haimrait (approved):

PR #403, opened 2021-01-06
fix(lambda.js): catch unhandled promise rejections

ronnathaniel:


src/wrappers/lambda.js
```diff
@@ -111,6 +114,17 @@ function baseLambdaWrapper(
...
         };
+        // catch an unhandled promise rejection within the user's code
+        process.once('unhandledRejection', (reason) => {
+            utils.debugLog('Unhandled Promise Rejection caught. Please handle the rejected promise');
+            const rejectedError = {
+                name: 'UnhandledPromiseRejectionError',
+                message: reason || '',
+                stack: '',
+            };
+            eventInterface.setException(runner, rejectedError, false);
```
henperez, line 122: Is this still needed?

ronnathaniel: If/when we add an earlier detection, this will set to show on dashboard

ronnathaniel: We can take it out now, but will have to be added in at that point

src/wrappers/lambda.js
```diff
@@ -10,7 +10,9 @@ const { getConfig } = require('../config.js');
 const awsLambdaTrigger = require('../triggers/aws_lambda.js');
 const eventInterface = require('../event.js');
 const lambdaRunner = require('../runners/aws_lambda.js');
-const { STEP_ID_NAME, MAX_VALUE_CHARS, EPSAGON_EVENT_ID_KEY } = require('../consts.js');
+const {
```
henperez: Not needed

src/wrappers/lambda.js
```diff
@@ -97,6 +99,7 @@ function baseLambdaWrapper(
             runner.setDuration(utils.createDurationTimestamp(startTime));
         };
 
+
```
henperez: Same

src/wrappers/lambda.js
```diff
@@ -295,6 +309,8 @@ function baseLambdaWrapper(
                         return returnValue;
                     });
             }
+
```
henperez: Same

henperez:
Overall looks good 💯  - Small changes

tvaintrob (approved):

PR #418, opened 2021-01-27
fix(tracer): optimizing filter trace

galbash:


src/tracer.js
```diff
@@ -462,19 +464,35 @@ module.exports.filterTrace = function filterTrace(traceObject, ignoredKeys, remo
         for (let i = 0; i < ignoredKeys.length; i += 1) {
             const predicate = ignoredKeys[i];
             if (typeof predicate === 'string' &&
-            config.processIgnoredKey(predicate) === config.processIgnoredKey(key)) {
+            predicate === config.processIgnoredKey(key)) {
```
galbash, line 467: moved the processing of the config to when the config is set (instead of each time)

src/config.js
```diff
@@ -174,8 +177,12 @@ module.exports.setConfig = function setConfig(configData) {
         config.disableHttpResponseBodyCapture = configData.disableHttpResponseBodyCapture;
     }
 
-    if (configData.ignoredKeys) {
-        config.ignoredKeys = configData.ignoredKeys;
+    if (configData.ignoredKeys && Array.isArray(configData.ignoredKeys)) {
+        const filteredIgnoredKeys = configData.ignoredKeys.filter(key => typeof key !== 'function');
```
amitlicht: consider filtering explicitly only for strings & regexes, and reject everything else with a warning (not just functions).

src/config.js
```diff
@@ -174,8 +177,12 @@ module.exports.setConfig = function setConfig(configData) {
...
+        if (filteredIgnoredKeys.length !== configData.ignoredKeys.length) {
+            console.warn('Epsagon deprecation warning: function ignoredKeys have been deprecated and will be ignored'); // eslint-disable-line no-console
```
amitlicht: I saw there's a `utils.printWarning` function, maybe use that? 
and in any case, add the problematic ignored keys to make this usable.

amitlicht (changes requested):

src/config.js
galbash: so the problem there is no clear way to filter a regexp (`typeof /wr32/ === 'object'`) - there are some hacks online but they are sketchy at best

src/config.js
```diff
@@ -174,8 +177,15 @@ module.exports.setConfig = function setConfig(configData) {
...
     }
 
-    if (configData.ignoredKeys) {
-        config.ignoredKeys = configData.ignoredKeys;
+    if (configData.ignoredKeys && Array.isArray(configData.ignoredKeys)) {
+        const filteredIgnoredKeys = configData.ignoredKeys.filter(key => typeof key !== 'function');
+        if (filteredIgnoredKeys.length !== configData.ignoredKeys.length) {
+            utils.printWarning(
+                'Epsagon deprecation warning: function ignoredKeys have been deprecated and will be ignored. received ignoredKeys:',
+                configData.ignoredKeys
```
amitlicht, line 186: not critical but I actually meant printing the specific invalid ignored keys.

amitlicht (approved):

PR #419, opened 2021-01-28
fix(simple-oauth2.js): add simple-oauth2 unit-test

ranrib:


enoodle:
@ranrib it is failing tests on node8 and node10

haimrait (dismissed):

PR #428, opened 2021-02-05
fix(wsk): fix hanging openwhisk actions

tripodsan:
I don't really understand why, but changing the `finally()` function from returning a value to a block solves our hanging actions problems.....maybe because returning a promise from the finally block makes it hanging but when it's returned into the same scope, it is not? 

maybe something todo with unresolved promises and node 12?

### DIFF
--- a/src/wrappers/openwhisk.js
+++ b/src/wrappers/openwhisk.js
@@ -129,11 +129,11 @@ function baseOpenWhiskWrapper(functionToWrap, options) {
                     eventInterface.setException(runner, err);
                     runnerSendUpdateHandler();
                     throw err;
-                }).finally(() => (
+                }).finally(() => {
                     tracer.sendTrace(runnerSendUpdateHandler).catch(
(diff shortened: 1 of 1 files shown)


enoodle:
@tripodsan Thanks for that, we are reviewing.

ranrib (approved):

enoodle (approved):

tripodsan:
BTW, I'm not sure if this was a good fix. it turns out, that there is different behaviour on macos and linux (I didn't test windows). 

with this new version on macos, the trace is sent before the response is returned.
but on linux, the trace is send after the response is returned. 

I'll do some more testing....

tripodsan:
@ranrib it would be cool, if someone from your team could investigate our _linux_ problem. it is easily reproducible here: https://github.com/adobe/helix-epsagon/issues/157

ranrib:
Yes we will investigate this

PR #429, opened 2021-02-05
fix(mongo_wrap_support): change mongo wrapper to be based on shimmer

haimrait:
Methods to wrap took from (https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/core/wireprotocol/index.js)
Support (insert, update, remove, query, command (count for example), getMore (For cursor pagination), 

Screen Shot 2021-02-05 at 17 58 30

### DIFF
--- a/src/events/mongodb.js
+++ b/src/events/mongodb.js
@@ -98,11 +98,11 @@ function getArgsFromFunction(...args) {
  * @returns {Array} Execiton of the called function
  */
 function internalMongodbOperationWrapper(...args) {
-    let patchedCallback;
     const relevantArgs = getArgsFromFunction(...args);
     const {
         server, namespace, cmd, callback, operationName, wrappedFunction,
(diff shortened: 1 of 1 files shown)


src/events/mongodb.js
```diff
@@ -1,105 +1,116 @@
...
 
 /**
- * Hook for mongodb requests starting
- * @param {Object} event The request event
+ * Wrap Mongodb operations call with tracing
+ * @returns {Array} Execiton of the called function
  */
-function onStartHook(event) {
+function internalMongodbOperationWrapper(...args) {
+    let patchedCallback;
```
galbash: initialize to `callback` otherwise we risk not calling the user callback in an exception in our instrumentation

galbash (changes requested):

galbash (approved):

enoodle (approved):

PR #431, opened 2021-02-10
fix(declaration): change labels declaration

ymwjbxxq:
On your documentation (https://docs.epsagon.com/docs/nodejs) you are reporting this
epsagon.init({
    token: <EPSAGON-TOKEN: string>,
    appName: <APP-NAME-STAGE: string>,
    labels: [ [<KEY>, <VALUE>], ['userID', userID] ]
});

When you are importing the library in a TypeScript project
import epsagon from 'epsagon';

you get this error: Type '[string, string]' is not assignable to type 'string'.

To fix the error and be able to initialize a label we need to change the definition in 
labels?: string[][]

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -11,7 +11,7 @@ declare module 'epsagon' {
     isEpsagonDisabled?: boolean
     urlPatternsToIgnore?: string[]
     ignoredKeys?: string[]
-    labels?: string[]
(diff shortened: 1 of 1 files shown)


ranrib:
Hi @ymwjbxxq - can you please open this PR from this repository? (you've been added as a collaborator).

PR #437, opened 2021-02-13
fix(ldap): ldapjs default to the first server of the array

ymwjbxxq:
ldapjs  v2.2.0 Add support for multiple servers (client)
https://github.com/ldapjs/node-ldapjs/releases

### DIFF
--- a/src/events/ldap.js
+++ b/src/events/ldap.js
@@ -29,7 +29,7 @@ function bindWrapper(bindFunction) {
             }
             utils.debugLog(`LDAP.js bind() wrapper - name: ${name}`);
             if (this.urls.length > 0) {
-              this.url = this.urls[0];
+                this.url = this.urls[0];
             }
             const resource = new serverlessEvent.Resource([
                 this.url.hostname,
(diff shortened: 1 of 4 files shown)


src/events/ldap.js
```diff
@@ -28,6 +28,9 @@ function bindWrapper(bindFunction) {
                 controls = c;
             }
             utils.debugLog(`LDAP.js bind() wrapper - name: ${name}`);
+            if (this.urls.length > 0) {
+                this.url = this.urls[0];
```
ranrib, line 32: Will this override `this.url` in some way?

ymwjbxxq: this.url is now undefined  because even do the documentation say you can set a string or array when
`const client = ldap.createClient({
        url: 'ldap://localhost:4389',
    });`
    
  the library set the .urls property while .url stay undefined and so we need to set it to avoid exceptions later on in your code.
  
I think this code is not 100% correct because the user could setup multiple servers for redundancy and so like this we pick only the first while maybe in case of errors we should try to pick other server in the array but at least in this way I did not change the logic.

ranrib:
Hi @ymwjbxxq - in order to make sure it works in both cases we'll use this PR and test it manually on our end. In the meantime I'm closing. Thanks!

PR #445, opened 2021-02-21
fix(sqs_utils.js): update sns-sqs support

ranrib:
Signature fields are not always coming through

### DIFF
--- a/src/resource_utils/sqs_utils.js
+++ b/src/resource_utils/sqs_utils.js
@@ -25,9 +25,7 @@ function getSNSTrigger(messages) {
                 'MessageId' in body &&
                 'TopicArn' in body &&
                 'Message' in body &&
-                'Timestamp' in body &&
-                'SignatureVersion' in body &&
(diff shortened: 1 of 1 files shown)


src/resource_utils/sqs_utils.js
```diff
@@ -25,9 +25,7 @@ function getSNSTrigger(messages) {
                 'MessageId' in body &&
                 'TopicArn' in body &&
                 'Message' in body &&
-                'Timestamp' in body &&
-                'SignatureVersion' in body &&
```
haimrait, line 29: Why did you remove them?
someone else uses them?

ranrib: This is used to identify the message within SQS got the SNS signature. in SNS FIFO both signature fields do not appear and that's why they are not crucial (TopicArn, MessageId, and the rest are a good indicator for SNS usage as well)

haimrait (approved):

PR #446, opened 2021-02-21
feat(ssm): ssm getParameter support

ymwjbxxq:
From the service map I noticed that SSM was not traced.

I tried to follow the same pattern that you inside the aws_sdk.js file for the SSM GetParameter function
https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameters.html

Not sure how do you test it but would be nice if you could let me know for future reference.


enoodle (approved):
Thanks Dan!

PR #458, opened 2021-02-25
fix(getmetricdata): add cloudwatch getMetricData event

ymwjbxxq:
I have added in the aws event the cloudwatch.getMetricData request
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudWatch.html#getMetricData-property


enoodle (approved):
Thanks Dan!

PR #460, opened 2021-03-06
fix(kinesis): add useful info for kinesis.putRecords trace

ymwjbxxq:
I changed the kinesis.putRecords section.

The request will report total_record_count and will add the payload when config.metadataOnly = false

In the response instead:
I removed shard_id and sequence_number because was picking up only the first and it is wrong in the PutRecords but I added:
total_record_count
kinesis_error_messages only if failed_record_count > 0

### DIFF
--- a/src/events/aws_sdk.js
+++ b/src/events/aws_sdk.js
@@ -179,7 +179,7 @@ const SNSEventCreator = {
 const SQSEventCreator = {
     /**
      * Updates an event with the appropriate fields from a SQS request
-     * @param {object} request The WS.Request object
+     * @param {object} request The AWS.Request object
      * @param {proto.event_pb.Event} event The event to update the data on
      */
     requestHandler(request, event) {
--- a/src/events/aws_sdk.js
+++ b/src/events/aws_sdk.js
@@ -117,6 +117,7 @@ const kinesisEventCreator = {
      */
     responseHandler(response, event) {
         let errorMessages = '';
+        let errorMessagesCount = 0;
         switch (response.request.operation) {
         case 'putRecord':
             eventInterface.addToMetadata(event, {
@@ -125,13 +126,14 @@ const kinesisEventCreator = {
             });
             break;
         case 'putRecords':
-            if (response.data?.FailedRecordCount > 0) {
+            if (response.data.FailedRecordCount && response.data.FailedRecordCount > 0) {
                 errorMessages = JSON.stringify(response.data.Records
                     .map(item => item.ErrorMessage));
+                errorMessagesCount = response.data.FailedRecordCount;
             }
             eventInterface.addToMetadata(event, {
                 total_record_count: `${response.data.Records.length}`,
-                failed_record_count: `${response.data?.FailedRecordCount || 0}`,
+                failed_record_count: `${errorMessagesCount}`,
                 kinesis_error_messages: errorMessages,
             });
             break;
--- a/src/events/aws_sdk.js
+++ b/src/events/aws_sdk.js
@@ -125,13 +125,13 @@ const kinesisEventCreator = {
             });
             break;
         case 'putRecords':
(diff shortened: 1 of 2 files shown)


src/events/aws_sdk.js
```diff
@@ -112,18 +116,23 @@ const kinesisEventCreator = {
...
-        case AWS_TYPES.kinesis.putRecord:
+        case 'putRecord':
             eventInterface.addToMetadata(event, {
                 shard_id: `${response.data.ShardId}`,
                 sequence_number: `${response.data.SequenceNumber}`,
             });
             break;
-        case AWS_TYPES.kinesis.putRecords:
+        case 'putRecords':
+            if (response.data.FailedRecordCount > 0) {
```
ranrib: Hi @ymwjbxxq did you test this part manually on your end? How does the `errorMessages` object look like for example?

src/events/aws_sdk.js
```diff
@@ -170,7 +179,7 @@ const SNSEventCreator = {
 const SQSEventCreator = {
     /**
      * Updates an event with the appropriate fields from a SQS request
-     * @param {object} request The AWS.Request object
+     * @param {object} request The WS.Request object
```
ranrib: Seems like you removed `A` by accident

ranrib:
Great PR @ymwjbxxq !

src/events/aws_sdk.js
ymwjbxxq: the output is 

```
export interface PutRecordsOutput {
    /**
     * The number of unsuccessfully processed records in a PutRecords request.
     */
    FailedRecordCount?: PositiveIntegerObject;
    /**
     * An array of successfully and unsuccessfully processed record results, correlated with the request by natural ordering. A record that is successfully added to a stream includes SequenceNumber and ShardId in the result. A record that fails to be added to a stream includes ErrorCode and ErrorMessage in the result.
     */
    Records: PutRecordsResultEntryList;
    /**
     * The encryption type used on the records. This parameter can be one of the following values:    NONE: Do not encrypt the records.    KMS: Use server-side encryption on the records using a customer-managed AWS KMS key.  
     */
    EncryptionType?: EncryptionType;
  }
```

and Records is made of

```
 export type ErrorMessage = string;
 export type PutRecordsRequestEntryList = PutRecordsRequestEntry[];
  export interface PutRecordsResultEntry {
    /**
     * The sequence number for an individual record result.
     */
    SequenceNumber?: SequenceNumber;
    /**
     * The shard ID for an individual record result.
     */
    ShardId?: ShardId;
    /**
     * The error code for an individual record result. ErrorCodes can be either ProvisionedThroughputExceededException or InternalFailure.
     */
    ErrorCode?: ErrorCode;
    /**
     * The error message for an individual record result. An ErrorCode value of ProvisionedThroughputExceededException has an error message that includes the account ID, stream name, and shard ID. An ErrorCode value of InternalFailure has the error message "Internal Service Failure".
     */
    ErrorMessage?: ErrorMessage;
  }
```

ymwjbxxq: @ranrib not sure how to make it fails but I can see in the trace
aws.kinesis.total_record_count

the failure should look like this:

```
{
    "FailedRecordCount": 2,
    "Records": [
        {
            "SequenceNumber": "49543463076548007577105092703039560359975228518395012686", 
            "ShardId": "shardId-000000000000"
        }, 
        {
            "ErrorCode": "ProvisionedThroughputExceededException",
            "ErrorMessage": "Rate exceeded for shard shardId-000000000001 in stream exampleStreamName under account 111111111111."
        },
        {
            "ErrorCode": "InternalFailure",
            "ErrorMessage": "Internal service failure."
        }
    ]
}
```

so the code should be fine.

ranrib (approved):

haimrait (approved):

PR #468, opened 2021-04-01
fix(performance): Require specific aws-sdk modules

roni-frantchi:
#### Issue
Requiring the entire `aws-sdk` seem to have noticeable performance penalty.  

#### Solution
Instead, require the global instance from `aws-sdk/global` or the needed client from `aws-sdk/clients/xxx`

### DIFF
--- a/test/acceptance/acceptance.js
+++ b/test/acceptance/acceptance.js
@@ -3,7 +3,7 @@
 const chai = require('chai');
 const chaiAsPromised = require('chai-as-promised');
 const { expect } = require('chai');
-const AWS = require('aws-sdk');
+const Lambda = require('aws-sdk/clients/lambda');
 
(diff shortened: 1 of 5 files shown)


test/acceptance/acceptance.js
```diff
@@ -3,7 +3,7 @@
 const chai = require('chai');
 const chaiAsPromised = require('chai-as-promised');
 const { expect } = require('chai');
-const AWS = require('aws-sdk');
+const Lambda = require('aws-sdk/clients/lambda');
```
roni-frantchi, line 6: This change is probably not required but 🤷‍♂️  
Might be a good idea to add an eslint rule that asserts no imports are made from `aws-sdk` - though it will not work for the various dynamic requires

roni-frantchi:
Addressed in #476

PR #473, opened 2021-04-08
feat: added try/catch around filterTrace call, added limit to recursion

tvaintrob:


src/tracer.js
```diff
@@ -412,10 +414,14 @@ function sendCurrentTrace(traceSender, tracerObject) {
...
-        Array.isArray(ignoredKeys) &&
-        ignoredKeys.length > 0 ?
-        module.exports.filterTrace(traceJson, ignoredKeys, removeIgnoredKeys) : traceJson;
+    try {
+        traceJson = ignoredKeys &&
+            Array.isArray(ignoredKeys) &&
+            ignoredKeys.length > 0 ?
+            module.exports.filterTrace(traceJson, ignoredKeys, removeIgnoredKeys) : traceJson;
+    } catch (err) {
+        utils.printWarning('Epsagon - failed to filter trace, cancelling send', err);
```
galbash, line 423: how is this canceling the send? isn't it just sending the unfiltered trace? (maybe you need to return at some point)

tvaintrob: fixed

haimrait (approved):

PR #486, opened 2021-05-12
feat(ssm): added support to SSM.getParameters, SSM.getParametersByPath

osherv:


tvaintrob (dismissed):

src/events/aws_sdk.js
```diff
@@ -923,9 +926,35 @@ const SSMEventCreator = {
...
+                );
+            }
+
+            if (response.data.Parameters && response.data.Parameters.length > 0) {
+                eventInterface.addToMetadata(event, {
+                    parameters: response.data.Parameters
+                        .map(singleParameter => ({
+                            Path: singleParameter.Path,
+                            Value: singleParameter.Name,
+                            Type: singleParameter.Type,
```
galbash: are we sure we want to capture the parameter value? I think it should be a secret

galbash: @ranrib I think you faced this question before

src/events/aws_sdk.js
```diff
@@ -923,9 +926,35 @@ const SSMEventCreator = {
...
+            if (response.data.InvalidParameters && response.data.InvalidParameters.length > 0) {
+                eventInterface.addToMetadata(
+                    event,
+                    { invalid_parameters: response.data.InvalidParameters }
+                );
+            }
+
+            if (response.data.Parameters && response.data.Parameters.length > 0) {
+                eventInterface.addToMetadata(event, {
+                    parameters: response.data.Parameters
```
galbash, line 949: maybe add it as a second dict to `addToMetadata` function, since this probably shouldn't be sent when metadataOnly = True (@idonava would love if you can give an osher a brief of `metadataOnly` mode)

src/events/aws_sdk.js
```diff
@@ -923,9 +926,35 @@ const SSMEventCreator = {
         switch (response.request.operation) {
         case 'getParameter':
             eventInterface.addToMetadata(event, {
-                value: (response.data.Parameter || { Value: '' }).Value || '',
+                parameter: {
```
galbash, line 929: same comment as below

src/events/aws_sdk.js
osherv: fixed

src/events/aws_sdk.js
ranrib: That should go to the non metadata section

osherv: fixed

src/events/aws_sdk.js
osherv: fixed

tvaintrob (approved):

src/events/aws_sdk.js
tripodsan: does this mean, that secret/encrypted parameters from the SSM are stored in the traces?

ranrib: Yes, but you can omit it with with the ignoredKeys (add parameters, or any key inside)

PR #487, opened 2021-05-16
feat(neo4j): added support to Neo4j operations

osherv:


package-lock.json
```diff
@@ -2364,6 +2364,14 @@
       "integrity": "sha1-YXmX/F9gV2iUxDX5QNgZ4TW4B2I=",
       "dev": true
     },
+    "ansi-align": {
```
haimrait: Why did you change the package-lock.json
Did you change the package.json?

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+const tracer = require('../tracer.js');
+const errorCode = require('../proto/error_code_pb.js');
+
+/**
+ * Extracts the relevant arguments from provided arguments
+ * @returns {Object} Object of the extracted arguments
+ */
+function getArgsFromFunction(...args) {
+    return {
+        query: args[0],
```
haimrait: what about transactionConfig?
https://neo4j.com/docs/api/javascript-driver/4.2/class/src/session.js~Session.html

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+    return { metadata, fullMetadata };
+}
+
+
+/**
+ * Wraps neo4j.Session.run and neo4j.Transaction.run functions with tracing
+ * @param {Function} wrappedFunction The function to wrap from Neo4j
+ * @returns {Function} The wrapped function
+ */
+function neo4jSessionRunWrapper(wrappedFunction) {
```
haimrait, line 271: Maybe spread it to two different wrappers that reuse the same code for better readability
like (neo4jSessionRunWrapper, neo4jTranssactionRunWrapper)

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+ * Parse from DB session with Neo4j server, useful metadata
+ * @param {neo4j.Session|neo4j.Transaction} session A DB session (could be Session of Transaction)
+ * @returns {{metadata: Object, fullMetadata: Object}} JSONs with command's metadata and command's
+ *      full metadata which contains additional data about the event
+ */
+function getNeo4JSessionMetadata(session) {
+    const metadata = {};
+    const fullMetadata = {};
+
+    let connectionHolder;
```
haimrait, line 98: Definition is not enough it will raise an exception if you don't init this value in case of default

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+        break;
+
+    default:
+        break;
+    }
+
+    metadata.dbName = connectionHolder._database ? connectionHolder._database : 'default';
+    metadata.operation = `${session.constructor.name}:${operationMode}`;
+
+    fullMetadata.config = connectionHolder._connectionProvider._config;
```
haimrait: add safe checks

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+}
+
+
+/**
+ * Gets session's address info from Transaction/Session instance
+ * @param {neo4j.Session|neo4j.Transaction} session A DB session (could be Session of Transaction)
+ * @returns {Object} json with address info { host, port }
+ */
+function getAddressInfo(session) {
+    let port = '7474';
```
haimrait: change host and port to consts

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+                'neo4j',
+                0,
+                errorCode.ErrorCode.OK,
+            ]);
+
+            dbApiEvent.setResource(resource);
+
+            eventInterface.addToMetadata(dbApiEvent, metadata, fullMetadata);
+
+            const resultResponse = wrappedFunction.apply(this, [query, params]);
```
haimrait: what about transactionConfig?

haimrait: https://neo4j.com/docs/api/javascript-driver/4.2/class/src/session.js~Session.html

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+                    // eslint-disable-next-line require-jsdoc
+                    onNext(record) {
+                        records.push(record);
+                        return observer.onNext.apply(this, arguments);
+                    },
+
+                    // eslint-disable-next-line require-jsdoc
+                    onCompleted(summary) {
+                        switch (summary.queryType) {
+                        case 's':
```
haimrait: why s is Read?

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+                        // eslint-disable-next-line prefer-rest-params
+                        return observer.onError.apply(this, arguments);
+                    },
+                });
+            };
+
+            return resultResponse;
+        } catch (error) {
+            tracer.addException(error);
+            return wrappedFunction.apply(this, [query, params]);
```
haimrait: what about transactionConfig?
https://neo4j.com/docs/api/javascript-driver/4.2/class/src/session.js~Session.html

src/events/neo4j.js
```diff
@@ -0,0 +1,265 @@
...
+
+            // Override the Result subscribe with patched subscriber
+            // eslint-disable-next-line func-names
+            resultResponse.subscribe = function (observer) {
+                const records = [];
+
+                return originalSubscribe.call(this, {
+                    ...observer,
+                    // eslint-disable-next-line no-unused-vars,require-jsdoc
+                    onKeys(_keys) {
```
haimrait: I think onKeys useless

src/events/neo4j.js
osherv: They can not be, they might be changed later this function.
should i change the function design?

src/events/neo4j.js
osherv: Fixed by added later this function 'if(connectionHandler)'
Is that the right way?
Otherwise i need to init a new connectionHandler instance, right?

src/events/neo4j.js
osherv: Fixed? (does that fix also fixed the prev comment?)

src/events/neo4j.js
haimrait: I meant change 7474 and neo4j to costs and not exact values

osherv: oh oh, fixed :O

src/events/neo4j.js
osherv: fixed! ty

src/events/neo4j.js
osherv: Fixed!

src/events/neo4j.js
osherv: Fixed!

src/events/neo4j.js
osherv: Finally Fixed XD

src/events/neo4j.js
osherv: Fixed

src/events/neo4j.js
osherv: Fixed

src/events/neo4j.js
```diff
@@ -0,0 +1,386 @@
...
+                    arguments,
+                    dbApiEvent,
+                    startTime
+                ));
+            };
+
+            return resultResponse;
+        } catch (error) {
+            tracer.addException(error);
+            return wrappedFunction.apply(this, [query, params]);
```
haimrait: I think if originalSubscribe.call will raise an exception we will call 
wrappedFunction.apply twice the first one before the subscribe function and the second inside the catch maybe we need to change the place of the catch scope

src/events/neo4j.js
```diff
@@ -0,0 +1,386 @@
...
+                    arguments,
+                    dbApiEvent,
+                    startTime
+                ));
+            };
+
+            return resultResponse;
+        } catch (error) {
+            tracer.addException(error);
+            return wrappedFunction.apply(this, [query, params, transactionConfig]);
```
haimrait: same as below

src/events/neo4j.js
haimrait: yea its ok

src/events/neo4j.js
```diff
@@ -0,0 +1,386 @@
...
+    session,
+    observer,
+    sessionArguments,
+    dbApiEvent,
+    startTime
+) {
+    const records = [];
+
+    return {
+        // eslint-disable-next-line require-jsdoc
```
haimrait, line 169: what about the spread of ...observer
why did you remove it?

haimrait (changes requested):

src/events/neo4j.js
osherv: By mistake, fixed, ty!

src/events/neo4j.js
osherv: Fixed by moving the apply out of the try scope, i think that will solve the issue, right?

src/events/neo4j.js
osherv: Fixed by moving the apply out of the try scope, i think that will solve the issue, right?

src/events/neo4j.js
```diff
@@ -0,0 +1,387 @@
...
+function neo4jTransactionRunWrapper(wrappedFunction) {
+    return function internalNeo4jTransactionRunWrapper(...args) {
+        const relevantArgs = getArgsFromFunction(...args, this.constructor.name);
+        const {
+            query, params,
+        } = relevantArgs;
+
+        utils.debugLog('User called Neo4j wrapped Transaction run function');
+
+        const resultResponse = wrappedFunction.apply(this, [query, params]);
```
haimrait: Now if the customer's code throws an exception we will not add it to the tracer object

src/events/neo4j.js
```diff
@@ -0,0 +1,387 @@
...
+function neo4jSessionRunWrapper(wrappedFunction) {
+    return function internalNeo4jSessionRunWrapper(...args) {
+        const relevantArgs = getArgsFromFunction(...args, this.constructor.name);
+        const {
+            query, params, transactionConfig,
+        } = relevantArgs;
+
+        utils.debugLog('User called Neo4j wrapped Session run function');
+
+        const resultResponse = wrappedFunction.apply(
```
haimrait: Now if the customer's code throws an exception we will not add it to the tracer object

src/events/neo4j.js
```diff
@@ -0,0 +1,387 @@
...
+        connectionHolder = session._connectionHolder;
+        break;
+
+    default:
+        return {
+            host, port,
+        };
+    }
+
+    if (session && connectionHolder && connectionHolder._connectionProvider) {
```
haimrait: why do you check the session here?

src/events/neo4j.js
osherv: fixed

src/events/neo4j.js
osherv: fixed

src/events/neo4j.js
osherv: right, fixed. thanks

haimrait (approved):

osherv:
bors merge

PR #498, opened 2021-06-20
feat(sql): ignore response rows from queries to specific tables

ronnathaniel:


src/events/sql.js
```diff
@@ -104,7 +104,10 @@ module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callbac
                         rows = res;
                     }
                     eventInterface.addToMetadata(dbapiEvent, { rowCount });
-                    if (rowCount && rows instanceof Array && rows.length) {
+                    if (rowCount &&
+                        rows instanceof Array &&
+                        rows.length &&
+                        !tracer.doesContainIgnoredKey(config.getConfig().ignoredDBTables, table)) {
```
ranrib: you've created a very big change just to check if `config.getConfig().ignoredDBTables.includes(table)`. Can you undo all changes including the packge-lock.json and simplify this PR please

ranrib (dismissed):

src/events/sql.js
```diff
@@ -50,9 +53,13 @@ module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callbac
             sqlObj.type = 'SQL-Command';
         }
 
-        const { type, table } = sqlObj;
+        const { type } = sqlObj;
+        let { table } = sqlObj;
+        if (!table) {
+            table = (sqlObj.from.length > 0 ? sqlObj.from[0].table : '');
```
haimrait: why did you change the table?
and if you got more than 1 table?
What happened if you select from two tables?

src/events/sql.js
```diff
@@ -30,11 +31,13 @@ module.exports.parseQueryArgs = function parseQueryArgs(arg1, arg2) {
...
  * @param {Array} params The params argument (values)
  * @param {Function} callback The callback argument (cb)
- * @param {Object} config The connection config object
+ * @param {Object} connConfig The connection config object
  * @param {string} driver The database driver type (mysql/pg/..)
  * @returns {Array} The arguments
  */
-module.exports.wrapSqlQuery = function wrapSqlQuery(queryString, params, callback, config, driver) {
+module.exports.wrapSqlQuery = function wrapSqlQuery(
+    queryString, params, callback, connConfig, driver
```
haimrait: better change connConfig to config

src/events/sql.js
```diff
@@ -5,6 +5,7 @@ const tracer = require('../tracer.js');
 const serverlessEvent = require('../proto/event_pb.js');
 const eventInterface = require('../event.js');
 const errorCode = require('../proto/error_code_pb.js');
+const config = require('../config.js');
```
haimrait: const config = require('../config.js'); change it to epsagon_config

src/events/sql.js
ronnathaniel: table extraction does not work for mysql2. Using mysql it comes from sqlObj->from[]->table
extraction now supports both if first extraction fails

src/events/sql.js
ronnathaniel: revert to epsagonConfig

src/events/sql.js
ronnathaniel: revert connConfig to config 👍

haimrait (approved):

PR #502, opened 2021-07-05
[GD-15] fixing lambda timeouts when filtering keys

tvaintrob:


src/trace_senders/http.js
```diff
@@ -19,7 +19,10 @@ const session = axios.create({
 
 module.exports.sendTrace = function sendTrace(traceObject) {
     utils.debugLog(`Posting trace to ${config.getConfig().traceCollectorURL}`);
-    utils.debugLog(`trace: ${JSON.stringify(traceObject, null, 2)}`);
+
+    if ((process.env.EPSAGON_DEBUG || '').toUpperCase() === 'TRUE') {
```
haimrait, line 23: why did you add the condition?
is the condition is not part of debugLog?

src/tracer.js
```diff
@@ -21,6 +21,7 @@ const logSender = require('./trace_senders/logs.js');
 const httpSender = require('./trace_senders/http.js');
 
 const FILTER_TRACE_MAX_DEPTH = 50;
+const MAX_EVENTS = 30;
```
haimrait: change it to be an env variable

src/trace_senders/http.js
tvaintrob: to avoid stringify when not in debug mode

src/tracer.js
tvaintrob: done

haimrait (dismissed):

haimrait (approved):

PR #503, opened 2021-07-06
feat(aws_sdk.js): support sqs batch messages

henperez:


tvaintrob (approved):

src/events/aws_sdk.js
```diff
@@ -230,6 +230,11 @@ const SQSEventCreator = {
             });
             break;
         case 'sendMessageBatch':
+            if (response.data.Successful && response.data.Successful.length > 0) {
```
haimrait, line 233: Check that response.data exists

src/events/aws_sdk.js
```diff
@@ -230,6 +230,11 @@ const SQSEventCreator = {
...
+                eventInterface.addToMetadata(event, {
+                    record: JSON.stringify(response.data.Successful.map(item => item)),
```
haimrait, line 235: what happened if json.stringify throws an error?

henperez: Internal exception that only fails the event

haimrait: where do you catch the exception?

src/events/aws_sdk.js
henperez: It can't be since it's part of the HTTP Response object (AWS SDK). We really want it to break in case of such a major API change rather than ignoring it. either way it'll be part of the Trace Exceptions[] and won't break the user code

PR #530, opened 2021-10-26
Sdk 158 sns v3

haddasbronfman:


src/events/aws_sdk_v3.js
```diff
@@ -0,0 +1,169 @@
...
+        }
+    },
+};
+
+/**
+ * a map between AWS resource names and their appropriate creator object.
+ */
+const specificEventCreators = {
+    sns: SNSv3EventCreator,
+};
```
osherv, line 60: New line is missing here

src/events/aws_sdk_v3.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+}
+
+module.exports = {
+    /**
+     * Initializes the @aws-sdk tracer
+     */
+    init() {
+        moduleUtils.patchModule(
+            '@aws-sdk/client-sns',
```
osherv, line 162: Why did u choose to add hook to client-sns and not to global?
In case we will want to add another Service support it will take much more time now instead of just add another event creator..

src/events/aws_sdk_v3.js
```diff
@@ -0,0 +1,169 @@
...
+        return cmd;
+    }
+}
+
+/**
+ * Wraps the @aws-sdk sns-client commands
+ * @param {Function} wrappedFunction The function to wrap
+ * @returns {Function} The wrapped function
+ */
+function AWSSDKv3WrapperSNS(wrappedFunction) {
```
osherv, line 82: Maybe i just missed it, but where r u capture the payloads?

src/events/aws_sdk_v3.js
haddasbronfman: Will open new issue for this

arbiv (approved):

src/events/aws_sdk_v3.js
haddasbronfman: currently for sns there no payloads

PR #534, opened 2021-11-07
fix(awssdkv3): increasing patch scope to catch more events

haddasbronfman:


nemoshlag:
I find "fix(awssdkv3): change instrumentation" very vague :) Can you elaborate a bit for documentation reason + it would help me to perform a review if I knew the purpose say:
fix(sdkv3-typo): fixing a typo OR fix(sdkv3-instrumentation): increasing patch scope to catch more events

src/events/aws_sdk_v3.js
```diff
@@ -79,11 +79,17 @@ function getOperationByCommand(command) {
...
-function AWSSDKv3WrapperSNS(wrappedFunction) {
-    return function internalAWSSDKv3WrapperSNS(command) {
+function AWSSDKv3Wrapper(wrappedFunction) {
+    return function internalAWSSDKv3Wrapper(command) {
         try {
-            const serviceIdentifier = this.config.serviceId.toLowerCase();// sns
+            const serviceIdentifier = this.config.serviceId.toLowerCase();
             const resourceName = '';
+
+            if (!(serviceIdentifier in specificEventCreators)) {
```
nemoshlag, line 88: this might be my personal taste but I feel that `if (!specificEventCreators.includes(serviceIdentifier)` is clearer. The specificEventsCreators need to be an array for that case. Also I would consider migrating the specificEventsCreators const to the consts file (maybe rename as well to expose it's SDK v3 events, say supportedSDKV3Events)

src/events/aws_sdk_v3.js
```diff
@@ -159,12 +165,10 @@ module.exports = {
      */
     init() {
         moduleUtils.patchModule(
-            '@aws-sdk/client-sns',
+            '@aws-sdk/smithy-client',
```
nemoshlag: pls add a comment for those stupid people like me who don't know what smithy-client is :)

src/events/aws_sdk_v3.js
haddasbronfman: It is the same as in v2. This is why I prefer to keep it like that.

osherv (approved):

src/events/aws_sdk_v3.js
```diff
@@ -165,7 +165,8 @@ module.exports = {
      */
     init() {
         moduleUtils.patchModule(
-            '@aws-sdk/smithy-client',
+            '@aws-sdk/smithy-client', // A client that can catch all 'send' commands
+            // sent from aws resources using aws-sdk v3.
```
arbiv, line 169: indentation seems weird... Is that intended?

arbiv (changes requested):

src/events/aws_sdk_v3.js
haddasbronfman: It is the same as in v2

src/events/aws_sdk_v3.js
nemoshlag: I would prefer having our supported events on a single js file instead of browsing each instrumentation

arbiv (approved):

PR #551, opened 2021-12-19
Fix/fix mongo ctx extraction

osherv:


src/events/mongodb.js
```diff
@@ -97,10 +97,16 @@ function getArgsFromFunction(...args) {
...
     if (operationName.includes('connection')) {
         const ctx = args[args.length - 3];
-        const hostParts = typeof ctx.address === 'string' ? ctx.address.split(':') : '';
-        const options = hostParts.length === 2 ?
-            { host: hostParts[0], port: hostParts[1] } :
-            {}; // no port means the address is a random UUID so no host either
+        let options = {};
+        if (ctx) {
+            const hostParts = typeof ctx.address === 'string' ? ctx.address.split(':') : '';
+            if (hostParts.length === 2) {
```
arbiv, line 103: if length != 2 options remains empty, and there no log. is that OK?

arbiv: @osherv yes

arbiv (approved):

PR #556, opened 2021-12-21
fix: add number and boolean types to label value

### LINKED ISSUE #550: Label TypeScript type definition
The current `index.ts` doesn't allow to use numeric or boolean labels but epsagon allows it.
Therefore I cannot sum or average my custom label.

I suggest changing the label definition to 
``` ts
  export function label(key: string, value: string | number | boolean): void
```

Thanks

guillaumeduboc:
fix: #550


guillaumeduboc:
The CI has failed because it hasn't found AWS credentials. Is it normal?

PR #561, opened 2022-01-30
Feat(AWS-SDK-V3): DynamoDB  CS-148

sagivr2020:


src/event.js
```diff
@@ -85,7 +85,7 @@ module.exports.addObjectToMetadata = function addObjectToMetadata(
     let objectToAdd = object;
     if (!config.getConfig().metadataOnly && dataFields.length > 0) {
         const fields = Object.getOwnPropertyNames(object).filter(
-            field => !dataFields.includes(field)
+            field => dataFields.includes(field)
```
nemoshlag, line 88: is this on purpose?

nemoshlag (approved):

src/event.js
sagivr2020: Yes, before the fix - it took all the fields that don't match the fields the function gets.

PR #563, opened 2022-02-03
feat(AWS-SDK-V3): support SQS V3

sagivr2020:


src/events/aws_sdk_v3.js
```diff
@@ -56,6 +57,71 @@ const SNSv3EventCreator = {
...
+     * @param {string} operation the operation we wrapped.
+     * @param {Command} command the wrapped command
+     * @param {proto.event_pb.Event} event The event to update the data on
+     */
+    requestHandler(operation, command, event) {
+        const parameters = command.input || {};
+        const resource = event.getResource();
+
+        if ('QueueUrl' in parameters) {
+            resource.setName(`${parameters.QueueUrl.split('/').pop()}`);
```
osherv: I think you should try catch in case of unexpected URL
Or add if Split('/') is not empty

src/events/aws_sdk_v3.js
```diff
@@ -56,6 +57,71 @@ const SNSv3EventCreator = {
...
+     * @param {string} operation the operation we wrapped.
+     * @param {object} response The AWS.Response object
+     * @param {proto.event_pb.Event} event The event to update the data on
+     */
+    responseHandler(operation, response, event) {
+        switch (operation) {
+        case 'SendMessageCommand':
+            eventInterface.addToMetadata(event, {
+                'Message ID': `${response.MessageId}`,
+                'MD5 Of Message Body': `${response.MD5OfMessageBody}`,
```
osherv, line 103: Is that the convention for MD5 tag?
Verify this

src/events/aws_sdk_v3.js
```diff
@@ -56,6 +57,71 @@ const SNSv3EventCreator = {
...
+            });
+            break;
+        case 'ReceiveMessageCommand': {
+            let messagesNumber = 0;
+            if (('Messages' in response) && (response.Messages.length > 0)) {
+                messagesNumber = response.Messages.length;
+                eventInterface.addToMetadata(event, {
+                    'Message ID': `${response.Messages[0].MessageId}`,
+                    'MD5 Of Message Body': `${response.Messages[0].MD5OfBody}`,
```
osherv, line 112: Same

src/events/aws_sdk_v3.js
```diff
@@ -56,6 +57,71 @@ const SNSv3EventCreator = {
...
+                });
+                const snsData = resourceUtils.getSNSTrigger(response.Messages);
+                if (snsData != null) {
+                    eventInterface.addToMetadata(event, { 'SNS Trigger': snsData });
+                }
+            }
+            eventInterface.addToMetadata(event, { 'Number Of Messages': messagesNumber });
```
osherv, line 119: Verify label name convention

src/events/aws_sdk_v3.js
```diff
@@ -56,6 +57,71 @@ const SNSv3EventCreator = {
     },
 };
 
+const SQSEventCreator = {
```
osherv: Should be SQSV3EventCreate i think

package.json
```diff
@@ -48,7 +48,7 @@
     "statements": 17,
     "branches": 13,
     "functions": 25,
-    "lines": 18
+    "lines": 17
```
osherv, line 51: Is that codecov or something?

src/events/aws_sdk_v3.js
sagivr2020: Yes, verified it

src/events/aws_sdk_v3.js
sagivr2020: Yes, also verified it

src/events/aws_sdk_v3.js
sagivr2020: Verified.

src/events/aws_sdk_v3.js
sagivr2020: Right, updated

package.json
sagivr2020: it shows in testing 17.99%, therefore I discussed with @nemoshlag to increase it.

src/events/aws_sdk_v3.js
sagivr2020: Updated to something like this:
if(data !== null && data !== '') {
   // do something
}

osherv (approved):

PR #571, opened 2022-02-16
fix(mem-leak): skip return value

nemoshlag:
Following memory leak issue -> CS-167 setting a more basic broad solution. Setting EPSAGON_SKIP_RETURN_VALUE to true will prevent lambda wrapper from documenting DB return value (which potentially can increase lambda's container memory usage)

### DIFF
--- a/src/wrappers/lambda.js
+++ b/src/wrappers/lambda.js
@@ -191,7 +191,7 @@ function baseLambdaWrapper(
                 eventInterface.addToMetadata(runner, { status_code: statusCode });
             }
             const config = getConfig();
-            if (error === null && !config.metadataOnly && !config.skipReturnVal) {
+            if (error === null && !config.metadataOnly && !config.skipReturnValue) {
                 try {
                     // Taken from AWS Lambda runtime
                     const jsonResult = JSON.stringify(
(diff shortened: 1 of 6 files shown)


test/unit_tests/test_config.js
```diff
@@ -102,6 +102,11 @@ describe('tracer config tests', () => {
         expect(config.getConfig()).to.contain({ decodeHTTP: false });
     });
 
+    it('setConfig: set skipReturnValue to true', () => {
```
arbiv: is it possible to add a functional test of this parameter?
(i.e return value is not added when this is true)

src/wrappers/lambda.js
```diff
@@ -190,11 +190,8 @@ function baseLambdaWrapper(
             if (statusCode) {
                 eventInterface.addToMetadata(runner, { status_code: statusCode });
             }
-
-            const table = utils.getObjectFromRunner(runner);
             const config = getConfig();
-
-            if (error === null && !config.metadataOnly && !config.ignoredDBTables.includes(table)) {
+            if (error === null && !config.metadataOnly && !config.skipReturnValue) {
```
arbiv: maybe returnValueEnabled will be better, to save us the headache of the double negation?

arbiv (changes requested):

arbiv (approved):

PR #573, opened 2022-03-01
Feat : integrate eventbridge for aws sdk v3

### LINKED ISSUE #572: Eventbridge operations for the AWS SDK V3 are not supported
Events put on an Eventbridge bus with the sdk v3 are not traced.

To compare the sdk v2 and v3 : 

```ts
import { APIGatewayProxyEvent } from 'aws-lambda';
import EventBridgeV2 from 'aws-sdk/clients/eventbridge';
import { EventBridge, PutEventsCommand } from '@aws-sdk/client-eventbridge';
import epsagon from 'epsagon';

const eventBridgeV2 = new EventBridgeV2();
const eventBridgeV3 = new EventBridge({});

epsagon.init({token, 'Demo-Epsagon'});

const demoEpsagon = async (event: APIGatewayProxyEvent) => {
  await eventBridgeV2
    .putEvents({
      Entries: [
        {
          Source: 'demo-epsagon-V2',
          DetailType: 'TEST',
          Detail: JSON.stringify({}),
          EventBusName: 'demo-epsagon',
        },
      ],
    })
    .promise();
  await eventBridgeV3.send(
    new PutEventsCommand({
      Entries: [
        {
          Source: 'demo-epsagon-V3',
          DetailType: 'TEST',
          Detail: JSON.stringify({}),
          EventBusName: 'demo-epsagon',
        },
      ],
    }),
  );
};

export const main = epsagon.lambdaWrapper(demoEpsagon);
```

guillaumeduboc:
Fix #572

### DIFF
--- a/src/events/aws_sdk_v3.js
+++ b/src/events/aws_sdk_v3.js
@@ -380,7 +380,7 @@ function AWSSDKv3Wrapper(wrappedFunction) {
 
             if (!(serviceIdentifier in specificEventCreators)) {
                 // resource is not supported yet
-                return wrappedFunction.apply(this, [command]);
+                return responsePromise;
             }
 
             const operation = getOperationByCommand(command);
--- a/src/events/aws_sdk_v3.js
+++ b/src/events/aws_sdk_v3.js
@@ -324,6 +324,7 @@ function getOperationByCommand(command) {
  */
 function AWSSDKv3Wrapper(wrappedFunction) {
     return function internalAWSSDKv3Wrapper(command) {
+        let responsePromise = wrappedFunction.apply(this, [command]);
         try {
             const serviceIdentifier = this.config.serviceId.toLowerCase();
             const resourceName = '';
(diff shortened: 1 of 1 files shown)


guillaumeduboc:
The promise returned by the aws-sdk v3 wrapper, was not the promise responsible for adding the event to the trace. 
Hence, adding the event to the trace was not awaited and the trace was often incomplete (`premature_exit: true`).

Furthermore, the wrapped function was called twice, resulting in 2 events published on the bus instead of one.


This bug is fixed in my first commit

src/events/aws_sdk_v3.js
```diff
@@ -324,6 +373,7 @@ function getOperationByCommand(command) {
  */
 function AWSSDKv3Wrapper(wrappedFunction) {
     return function internalAWSSDKv3Wrapper(command) {
+        let responsePromise = wrappedFunction.apply(this, [command]);
```
galbash, line 376: I this this may lead to us calling the function twice for unsupported services

guillaumeduboc: Nice spot, I've fixed it in the next commit

PR #574, opened 2022-03-03
ci: added dependabot auto-merge github action

arbiv:


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
+          github-token: ${{ secrets.mytoken }}
```
haddasbronfman: To my understanding we should create secret in this repo according to this guide: 
https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository

and use it like we use other secrets (for example, here: https://github.com/epsagon/otel-js/blob/6a9a59ddbddf54d4d0468199b0c4358edc57f284/.github/workflows/release.yaml#L23)

arbiv: Good catch! Fixed. Changed it to GITHUB_TOKEN that is generated automatically (https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)

nemoshlag (approved):

nemoshlag (approved):

PR #581, opened 2022-03-27
Fix/handing stripe call version above 1.169.0/CS-174

sagivr2020:


src/events/http.js
```diff
@@ -174,6 +174,20 @@ function httpWrapper(wrappedFunction) {
...
+            // Skipping new stripe calls since it interfere with async events
+            if (options.headers['User-Agent']) {
+                if (options.headers['User-Agent'].includes('Stripe/v1 NodeBindings/')) {
+                    try {
+                        const stripeVersion = parseInt(options.headers['User-Agent'].split('/')[2].split('.')[1], 10);
+                        if (stripeVersion > 169) {
+                            return wrappedFunction.apply(this, [a, b, c]);
+                        }
+                    } catch (err) {
+                        utils.debugLog('Could not catch stripe call');
```
arbiv: I'm not sure that the error message here describes the actual error

src/events/http.js
```diff
@@ -174,6 +174,20 @@ function httpWrapper(wrappedFunction) {
                 return wrappedFunction.apply(this, [a, b, c]);
             }
 
+            // Skipping new stripe calls since it interfere with async events
+            if (options.headers['User-Agent']) {
+                if (options.headers['User-Agent'].includes('Stripe/v1 NodeBindings/')) {
+                    try {
+                        const stripeVersion = parseInt(options.headers['User-Agent'].split('/')[2].split('.')[1], 10);
+                        if (stripeVersion > 169) {
+                            return wrappedFunction.apply(this, [a, b, c]);
```
arbiv: Should the return be also wrapped in try-catch, or only the parseInt?
In other cases in this (unbelievingly long) function the return is not wrapped

arbiv (changes requested):

arbiv (approved):

PR #592, opened 2022-06-07
feat: override status code with thrown error message 

ronnathaniel:


nemoshlag (dismissed):

src/wrappers/lambda.js
```diff
@@ -178,7 +178,13 @@ function baseLambdaWrapper(
                         message: errorMessage,
                     };
                 }
-
+                // Attempt to extract Status Code from err message.
+                const parsedErrMessage = utils.parseRelaxedJSON(reportedError.message);
```
ivojinovic: Hey I'm sorry, I think I mislead you with what I wrote in the email: not all of our error messages are stringifyed JSON with `statusCode` - in fact a lot of them are just a string containing `400` or `500`. So, I think it would be best if this worked exactly like the mapping in or APIGW IntegrationResponses: by searching the string for a pattern like this `.*50[0-9].*`  or `.*40[0-9].*`. Ideally, it would be able to capture the exact code (such as `403`, or `502` if need be), because sometimes the mappings look for those specific codes and we may want in our traces exactly what the FE will receive

ivojinovic (approved):

nemoshlag (approved):

PR #608, opened 2023-01-02
Fix/adding session constructor name for neo4j version 4/cs 000

sagivr2020:
Got a case from Syngenta related to Neo4j version 4: https://epsagon.slack.com/archives/C01JRMKDPB9/p1669754592992149


sagivr2020:
Another PR has been opened regarding the same issue.

